PR #10, opened 2018-02-22
SOOP CO2 pipeline V2- Processing of SOOP_CO2 DM, RT and FutureReefMap…

bpasquer:
… data


bpasquer:
I wanted to submit file related to my piupeline only but ended up adding one unwanted file. So please ignore   setup.py

mhidas:
Ok, but we'll need to add that file back in, otherwise it breaks _everything_, including the Travis tests.

setup_virtualenv.sh
```diff
@@ -29,10 +29,10 @@ function setup_virtualenv() {
...
 
     echo "Installing dependencies into virtual environment..."
-    ${VIRTUALENV_PIP} install --quiet --upgrade ${WHEEL_CACHE_DIR}/compliance_checker-*.whl
-    ${VIRTUALENV_PIP} install --quiet -c constraints.txt --upgrade ${WHEEL_CACHE_DIR}/cc_plugin_imos-*.whl
-    ${VIRTUALENV_PIP} install --quiet -c constraints.txt --upgrade ${WHEEL_CACHE_DIR}/aodncore-*.whl
-    ${VIRTUALENV_PIP} install --quiet -c constraints.txt -r requirements.txt
+    ${VIRTUALENV_PIP} install  --upgrade ${WHEEL_CACHE_DIR}/compliance_checker-*.whl
+    ${VIRTUALENV_PIP} install -c constraints.txt --upgrade ${WHEEL_CACHE_DIR}/cc_plugin_imos-*.whl
+    ${VIRTUALENV_PIP} install -c constraints.txt --upgrade ${WHEEL_CACHE_DIR}/aodncore-*.whl
+    ${VIRTUALENV_PIP} install -c constraints.txt -r requirements.txt
```
mhidas: Did you mean to commit this as well? If so it should be in a separate PR.

bpasquer:
I've put back the deleted file . and pushed 2 other that may be useful. But Travis is  still unhappy. He 's never been properly introduced to me  and already making my life harder....

setup_virtualenv.sh
bpasquer: ok will do

mhidas:
@bpasquer, meet [Travis](https://travis-ci.org/), and vice versa.
Travis automatically runs a vm, sets up the required environment, and runs all the unittests in the repo, every time a new commit is pushed to GitHub, or a PR is submitted, then tells GitHub about the outcome.

See the message at the bottom that says "All checks have passed"?
That means Travis is happy! :tada:

bpasquer:
Then nice to meet you Travis!

aodndata/soop/soop_co2_classifier.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if src_file.name.endswith('.nc'):
+            fields = cls._get_file_name_fields(src_file.name, min_fields=5)
+            if 'FV00' in src_file.name:
+                return fields[1] + '_RT'  # <Facility-Name + RTnc suffix>
+            else:
+                return fields[0]  # <Project-Name>
+        else:
+            raise InvalidFileFormatError(
+                "'{name}'is not a valid RT CO2 data (*dat.txt) or not a NetCDF file.".format(
```
mhidas: So far this is only handling '.nc' files. Is it also supposed to handle '*dat.txt'?

aodndata/soop/soop_co2_classifier.py
```diff
@@ -0,0 +1,129 @@
...
+
+            dir_list.append('Future_Reef_MAP')
+            data_type = 'underway'
+            dir_list.extend([data_type, ship_callsign(ship_code)])
+
+        if project in ['IMOS', 'FutureReefMap']:
+            att_list = cls._get_nc_att(src_file.src_path, ['cruise_id', 'time_coverage_start'])
+            year = att_list[1][:4]
+            cruise_id = att_list[0]
+            if not cruise_id:
```
mhidas: You don't need to check for this. If the attribute is missing, the same exception is raised in the `_get_nc_att()` method.

aodndata/soop/soop_co2_classifier.py
```diff
@@ -0,0 +1,129 @@
+from aodncore.pipeline import FileClassifier
+from aodncore.pipeline.exceptions import InvalidFileNameError, InvalidFileContentError, InvalidFileFormatError
+from ship_callsign import ship_callsign_list, ship_callsign
+import datetime
+
+VALID_PROJECT = ['IMOS', 'FutureReefMap', 'SOOP-CO2_RT']
+VESSEL_CODE = {'AA': 'VNAA',
+               'IN': 'VLMJ'}
```
mhidas: Since all your code is conveniently packaged in a class, you could move these global variables into the class as well:
```python
class SoopCo2FileClassifier(FileClassifier):
    VALID_PROJECT = ['IMOS', 'FutureReefMap', 'SOOP-CO2_RT']
    VESSEL_CODE = {'AA': 'VNAA', 'IN': 'VLMJ'}
```
Then you can access them within the methods as e.g. `cls.VALID_PROJECT`

setup_virtualenv.sh
lwgordonimos: Yeah don't trash this file please... it's like getting your car back from the mechanic missing the steering wheel ;)

aodndata/soop/soop_co2_classifier.py
bpasquer: it is handling the 'dat.txt' above

aodndata/soop/soop_co2_nrt_nc_generator.py
```diff
@@ -0,0 +1,355 @@
...
+    """
+    Read in data from co2 realtime file and produce a netcdf file
+    """
+    # Parse data into dataframe
+    (dataf, platform_code) = read_realtime_file(self)
+    # format data
+    (dtime, time) = get_time_formatted(dataf)
+    # generate nc file name
+    netcdf_filename = create_netcdf_filename(platform_code, dtime)
+    tmpdir = mkdtemp()
```
lwgordonimos: This should be the products_dir from the handler instead of just a random temp dir. This helps make sure that the files all get created under the same temp dir for the handler.

You could update the function to accept as an argument, and then pass through "self.products_dir" from the handler when you call it, e.g.

```python
def process_co2_rt(realtime_file_path, temp_dir):
```

setup.py
```diff
@@ -23,6 +23,7 @@
         'SrsGhrsstHanlder = aodndata.srs.srs_ghrsst:SrsGhrsstHandler',
         'SrsOcBodBawHandler = aodndata.srs.srs_oc_bodbaw:SrsOcBodBawHandler',
         'SrsOcLjcoHandler = aodndata.srs.srs_oc_ljco:SrsOcLjcoHandler'
+        'SoopCo2Handler = aodn.soop.soop_co2:SoopCo2Handler'
```
lwgordonimos: Should be "aodndata.soop.soop_co2:SoopCo2Handler" ?

test_aodndata/soop/test_soop_co2.py
```diff
@@ -0,0 +1,103 @@
...
+            self.assertEqual(f.dest_path,
+                             'IMOS/SOOP/SOOP-CO2/VNAA_Aurora-Australis/2017/AA1617_V3/' + f.name)
+            self.assertEqual(f.check_type, PipelineFileCheckType.NONEMPTY_CHECK)
+            self.assertEqual(f.publish_type, PipelineFilePublishType.UPLOAD_ONLY)
+
+    def test_good_file_with_compliance_check(self):
+        # we also expect this to succeed, since the test file is known be CF compliant
+        self.run_handler(GOOD_NC, check_params={'checks': ['cf']})
+
+        if __name__ == '__main__':
```
lwgordonimos: This is indented so that it is inside the test method, should be in the module root (i.e unindented).

aodndata/soop/soop_co2.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # first process the NetCDF file to set the destination path for the file collection
+            nc = nc_file[0]
+            set_nc_dest_path = SoopCo2FileClassifier.dest_path(nc)
+            nc.dest_path = os.path.join(set_nc_dest_path, nc.name)
+            nc.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
+            nc.check_type = PipelineFileCheckType.NC_COMPLIANCE_CHECK
+
+            # SOOP-CO2 DM and FRMAP .txt,.pdf or/and .xml files. Set check type to NONEMPTY and publish type to UPLOAD_ONLY
+            non_nc_files = self.file_collection.filter_by_attribute_id('check_type', PipelineFileCheckType.NO_ACTION)
```
lwgordonimos: The "check type" isn't the best way to differentiate files by type. There isn't a specific "filter" method for this exact use case, but you could also get a collection of non-NetCDF files like this:

```python
non_nc_files = PipelineFileCollection(f for f in self.file_collection if f.file_type is not FileType.NETCDF)
```

aodndata/soop/soop_co2.py
```diff
@@ -0,0 +1,80 @@
...
+            for non_nc in non_nc_files:
+                non_nc.check_type = PipelineFileCheckType.FORMAT_CHECK
+                non_nc.publish_type = PipelineFilePublishType.UPLOAD_ONLY
+                non_nc.dest_path = os.path.join(set_nc_dest_path, non_nc.name)
+
+        elif self.input_file.endswith('dat.txt'):
```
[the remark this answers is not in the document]
bpasquer: ```You don't need to assign the dest_path attribute to the file or call the dest_path function yourself. This should be calculated and assigned automatically later on during the publish step, based on the fact that you added the following below```
I call the dest_path function in case when I need to work out the destination path for ancillary files that don't carry any/enough information for the dest_path function to calculate it directly.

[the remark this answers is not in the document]
bpasquer: and Thanks for your input @lwgordonimos  👍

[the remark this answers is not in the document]
bpasquer: '''The dest_path function/method should return the entire path, including the filename.'''
Except that it's not convenient to re-use the dest path if it includes the (netcdf) filename.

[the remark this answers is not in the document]
lwgordonimos: Ah yes, that rings a bell. Makes sense. I have been thinking a bit about the dest_path stuff and think this sort of thing could be made cleaner by:

1. being able to add some custom metadata to a PipelineFile
1. making the dest_path functions take a PipelineFile object as a parameter (instead of just a path)

This would mean the dest_path functions would potentially have a lot more information to work with. But definitely a "future enhancement".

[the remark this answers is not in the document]
bpasquer: Are you saying I should leave the bits where I process ancillary files off the pipeline until the enhancement is implemented?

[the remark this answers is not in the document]
lwgordonimos: Sorry, wasn't clear. No, I'm saying to do exactly what you're doing here now, because it's the only (and currently best) way, but that future enhancement idea could make this simpler.

This way is perfectly valid, and is in fact why I added this check https://github.com/aodn/python-aodncore/blob/master/aodncore/pipeline/files.py#L693, so that if a handler author wanted to set it manually, the publish step would *not* try to overwrite it, because it assumes it was set earlier  for a reason :)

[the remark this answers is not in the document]
bpasquer: '''PipelineFilePublishType.HARVEST_UPLOAD and PipelineFileCheckType.NC_COMPLIANCE_CHECK are the defaults for NetCDF files, so you don't need to explicitly assign them here'''
Is this suppose to apply also to product? As soon as you create a netcdf file it these attribute are assigned by default? Cause this is not the behaviour i see when running the unittest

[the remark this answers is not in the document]
lwgordonimos: No you're right, not for products. By default they are NO_ACTION, so you can set them to whatever.

I should have put the comment against the actual line:
https://github.com/aodn/python-aodndata/pull/10/files/5d95d90e833eaccc919f6cbf3ce1086fa1966304#diff-c9a870baa45ba537fc28ab45791fdc2bR69

PR #13, opened 2018-03-05
Add basic validation for package entry points. Test for common 'missi…

lwgordonimos:
…ng comma' silent error and test that referenced object is resolvable.


anguss00:
hmm not sure why Trav is complaining... just looks like the push test stalled

PR #14, opened 2018-03-05
Update .travis.yml

lwgordonimos:
Similar to https://github.com/aodn/python-aodncore/pull/54


lwgordonimos:
@mhidas I got rid of the compliance checker, and looks like everything else is now up to speed in order to make it work. Should be good to go now!

PR #16, opened 2018-03-08
setup_virtualenv script to install core checker from PyPI

mhidas:
Core checker version defined by requirements in IMOS plugin & aodncore.


mhidas:
@lwgordonimos The Travis checks only failed because of the issues resolved in #17. Anyway I don't think any of the changes here are tested by Travis, so should be safe to ignore. Otherwise we can rebase this branch once #17 is merged.

PR #28, opened 2018-05-10
SRS OC LJCO - update handler to deal with DALEC data

lbesnard:


aodndata/srs/srs_oc_ljco.py
```diff
@@ -23,24 +23,39 @@ def dest_path(self, filepath):
                       netcdf_filename)
 
         if m is None:
-            raise ValueError("file name not matching regex to deduce dest_path")
+            m = re.search('^IMOS_SRS-OC-LJCO_(.*)_(.*)_LJCO_FV(.*)\.nc$', netcdf_filename)
```
lwgordonimos: I think this whole regex matching/extracting bit can be cleaned up a lot. There are so many regex operations which look unnecessary, and I think the use of "named groups" would enhance readability a lot, and get rid of a lot of lines (remember, more lines = more bugs!).

test_aodndata/srs/test_srs_oc_ljco.py
```diff
@@ -30,6 +33,11 @@ def test_netcdf(self):
         self.assertTrue(f.is_checked)
         self.assertTrue(f.is_stored)
 
+    def test_netcdf_dalec(self):
```
lwgordonimos: We need some more tests for these dest path functions, otherwise they don't catch problems (which is the whole point of the tests really!). 

Each function should have at least:
1) known good path (i.e. should succeed)
2) at least deliberate 2 failures, or more depending on how complex the function is in determined the name

The failures should be, for example, 1) a file just with a completely wrong naming scheme, 2) a file which looks mostly correct but has invalid date / product name etc. etc. 3) if it opens the file, a file that is invalid NetCDF etc.

The failures are actually the *main* thing to test, to pick up problems before they are packaged, so in general, you should go back through your tests and make sure there are more strict tests for these functions, because if these are well tested, it's very unlikely that dest path related problems come up when it is actually deployed.

For example, I think this sort of thing would be easily picked up in testing, rather than after it's already been deployed: https://github.com/aodn/python-aodndata/issues/27

PR #31, opened 2018-05-17
implicit version install of netcdf4 to avoid netcdftime issue

anguss00:
@lwgordonimos thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
 install:
   - pip install cc_plugin_imos-*.whl
   - pip install aodncore-*.whl
+  - pip install NetCDF4==1.3.1
   - pip install -r requirements.txt
 
 script:


mhidas:
Will https://github.com/aodn/cc-plugin-imos/pull/12, combined with the chain of dependencies (aodndata > aodncore > cc-plugin-imos) ensure that the right version of netCDF4 will be used everywhere?
(That is, once we build and push the updated cc-plugin-imos to the prod environment. I've already built it on Jenkins.)

anguss00:
Probably yes, this will then be an unnecessary step

anguss00:
The magic of module dependency management is a black art to me ;-)

mhidas:
Passed!
@lbesnard All your other PRs should also pass now, if this was the only thing wrong with them.

PR #37, opened 2018-05-23
Argo - add regex to handler. Improve unittesting of argo handler

lbesnard:


lbesnard:
need to promote aodncore new version before re starting travis

lbesnard:
@lwgordonimos this now passes travis

mhidas:
I just ran all the unittests on your branch, and the Argo tests passed with aodncore 0.10.2 (current prod) and aodncore 0.11 (current RC), but failed with the previous 0.10.1
Should we start specifying a minimum version in the requirements file?

lbesnard:
yeap this is normal. Something on core had to be modified. I think core and data are quite linked, both need the latest version of each other so everything can run smoothly. So I feel like adding a specific version in requirements.txt would not make things any easier

test_aodndata/argo/test_argoHandler.py
```diff
@@ -19,6 +23,28 @@ def test_rsync_manifest_file(self):
...
+        handler.resolve_params = {'relative_path_root': os.path.join(TEST_ROOT, 'Argo/dac')}
+        handler.run()
+
+        self.assertIsInstance(handler.error, MissingFileError)
+
+    def test_non_nc_rsync_manifest_file(self):
+        # check file not finishing with nc extension is not pusblished
+        handler = self.handler_class(NON_NC_RSYNC_MANIFEST_FILE)
+        handler.resolve_params = {'relative_path_root': os.path.join(TEST_ROOT, 'Argo/dac')}
+        handler.run()
```
bpasquer: According to Leigh, it's better to use the run_handler class 
`
handler = self.run_handler()
`

lwgordonimos: Yes! The reason for this is that the handler does *not* raise an exception in case of an error, so it can hide bugs and errors if the test is not explicitly looking for it.

run_handler calls the run method and then checks the "handler.error" attribute is None, otherwise it fails the test.

https://aodn.github.io/python-aodncore/_modules/aodncore/testlib/handlertest.html#HandlerTestCase.run_handler

lbesnard: @lwgordonimos I have modified the code to add some preprocess. also made the unittests more explicit

aodndata/argo/handler.py
```diff
@@ -10,6 +10,14 @@ def __init__(self, *args, **kwargs):
 
         relative_path_root = os.path.join(self._config.pipeline_config['global']['wip_dir'], 'Argo/dac')
         self.resolve_params = {'relative_path_root': relative_path_root}
+        self.include_regexes = ['.*\.nc']
+
+    def preprocess(self):
+        """ set NO_ACTION to non NetCDF files"""
+        self.file_collection.set_publish_types_from_regexes(addition_type=PipelineFilePublishType.NO_ACTION,
```
lwgordonimos: This doesn't read very intuitively to me, since it's kind of inverting the intent of that method, which is a bit clunky and specific (I wrote it early on in the project, and now I look at it it's a bit quirky). I'd be more explicit with something like the filter and set methods, which are nicer, e.g.:

```python
        self.file_collection.\
            filter_by_attribute_id_not('file_type', FileType.NETCDF).\
            set_publish_types(PipelineFilePublishType.NO_ACTION)
```

With the file collection, filter it by non-NetCDF files and set them to NO_ACTION.

test_aodndata/argo/test_argoHandler.py
lwgordonimos: "test_non_nc_rsync_manifest_file" doesn't have an assertion for handler.error? This should *always* assert either None if no error expected, or the specific Exception class if one is expected.

lbesnard:
@lwgordonimos  done

PR #41, opened 2018-05-28
Call PipelineFile with option is_deletion for previous version files

bpasquer:


lwgordonimos:
This will likely work, but I think for these handlers that delete previous versions, we really need some tests to confirm the "previous version" handling code. We need to have a higher degree of confidence that they behave as expected before pushing to RC for testing, since we now have the packaging process in between GitHub and the server.

For example, have a test here: https://github.com/aodn/python-aodndata/blob/master/test_aodndata/soop/test_soop_ba.py#L21 create a basic structure of keys in the "upload_uri" location to represent the previous versions and some other random files, so that you can test which ones are deleted, and confirm that it works as expected. Shouldn't be too hard, happy to assist!

lbesnard:
@lwgordonimos agree. Probably need to force in which sub-dir of the bucket objects are allowed to be deleted

lwgordonimos:
I just mocked this up to show how I'd probably go about artificially creating an existing S3 state in order to test the previous version code:

```python
    def test_setup_upload_location(self):
        # create some PipelineFiles to represent the existing files on 'S3'
        preexisting_files = PipelineFileCollection()
        existing_file1 = PipelineFile(GOOD_NC, dest_path='IMOS/my/temporary/path1')
        existing_file2 = PipelineFile(BAD_NC, dest_path='IMOS/my/temporary/path2')
        preexisting_files.update((existing_file1, existing_file2))

        # set the files to UPLOAD_ONLY
        preexisting_files.set_default_publish_types(include_regexes='.*', exclude_regexes=None,
                                                    addition_type=PipelineFilePublishType.UPLOAD_ONLY,
                                                    deletion_type=PipelineFilePublishType.NO_ACTION)

        # upload the 'preexisting_files' collection to the unit test's temporary upload location
        broker = get_storage_broker(self.config.pipeline_config['global']['upload_uri'])
        broker.upload(preexisting_files)

        # run the handler
        handler = self.run_handler(GOOD_ZIP)

        # add some tests to make sure the previous files were handled appropriately, e.g.
        # - they were added as deletions
        # - they were successfully deleted
        # - they were the *only* ones deleted
        pass
```

PR #43, opened 2018-05-30
ACORN pipeline 2 - basic working destpath for both RT and DM

lbesnard:
WIP only
@ggalibert FYI. Next step is to add a check to know what is on S3


aodndata/acorn/handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class AcornHandler(HandlerBase):
+    def __init__(self, *args, **kwargs):
+        super(AcornHandler, self).__init__(*args, **kwargs)
+        self.allowed_extensions = ['.nc']
+
+    @staticmethod
+    def dest_path(filepath):
+        m = re.search(r'^IMOS_ACORN_([A-Z].*)_([0-9]{8}T[0-9]{6}Z)_([A-Z].*)_(FV0[0-2])_(.*)\.nc$', filepath)
```
ggalibert: Original bash regex is as follow:
```Bash
ACORN_RT_REGEX='^IMOS_ACORN_[[:alpha:]]{1,2}_[[:digit:]]{8}T[[:digit:]]{6}Z_[[:alpha:]]{3,4}_FV00_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$'
ACORN_DM_REGEX='^IMOS_ACORN_[[:alpha:]]{1,2}_[[:digit:]]{8}T[[:digit:]]{6}Z_[[:alpha:]]{3,4}_FV01_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$'
```

Can we keep the constraint on the site / station code limited to 3 or 4 characters? Also there is no FV02 allowed, only FV00 and FV01.

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,69 @@
...
+        if m is None:
+            raise InvalidFileNameError(
+                "file name: \"{filename}\" not matching regex to deduce dest_path for acorn".format(
+                    filename=os.path.basename(filepath)))
+
+        station_name = m.group(3)
```
ggalibert: Can either be a station or a site code. Let's call this variable `platform_code`

setup.py
```diff
@@ -10,6 +10,7 @@
 ENTRY_POINTS = {
     'pipeline.handlers': [
         'AatamsNrtHandler = aodndata.aatams.aatams_nrt:AatamsNrtHandler',
+        'AcornHandler = aodndata.acordn.handler:AcornHandler',
```
ggalibert: We need 2 distinct pipelines. One for RT and one for DM. The reason is DM files can be uploaded per thousands and take time to be processed, so in the meantime we don't want the RT flow to be affected.

lwgordonimos: Just chiming in on this one... this package only provides the *handler*, and so this package only needs to worry about that. The split directories for RT and DM and the different regexes should be reflected in the *pipeline* configuration rather than the handler definition here.

We'll need to look at how to most cleanly map the existing RT/DM over, which won't be too hard.

test_aodndata/acorn/test_acorn_dm.py
```diff
@@ -0,0 +1,54 @@
...
+    'IMOS_ACORN_RV_20161121T003000Z_GUI_FV00_radial.nc': 'IMOS/ACORN/radial/GUI/2016/11/21',
+    'IMOS_ACORN_RV_20171014T060000Z_LANC_FV00_radial.nc': 'IMOS/ACORN/radial/LANC/2017/10/14',
+    'IMOS_ACORN_RV_20180907T210500Z_LEI_FV00_radial.nc': 'IMOS/ACORN/radial/LEI/2018/09/07',
+    'IMOS_ACORN_RV_20190825T053500Z_NNB_FV00_radial.nc': 'IMOS/ACORN/radial/NNB/2019/08/25',
+    'IMOS_ACORN_RV_20200718T140000Z_NOCR_FV00_radial.nc': 'IMOS/ACORN/radial/NOCR/2020/07/18',
+    'IMOS_ACORN_RV_20210611T085000Z_RRK_FV00_radial.nc': 'IMOS/ACORN/radial/RRK/2021/06/11',
+    'IMOS_ACORN_RV_20220504T010000Z_SBRD_FV00_radial.nc': 'IMOS/ACORN/radial/SBRD/2022/05/04',
+    'IMOS_ACORN_RV_20230401T012000Z_TAN_FV00_radial.nc': 'IMOS/ACORN/radial/TAN/2023/04/01',
+    'IMOS_ACORN_V_20180910T010000Z_BONC_FV00_sea-state.nc': 'IMOS/ACORN/vector/BONC/2018/09/10',
+    'IMOS_ACORN_V_20140804T010000Z_TURQ_FV00_sea-state.nc': 'IMOS/ACORN/vector/TURQ/2014/08/04',
```
ggalibert: Let's remove the FV00 examples above.

test_aodndata/acorn/test_acorn_rt.py
```diff
@@ -0,0 +1,56 @@
...
+    'IMOS_ACORN_RV_20181113T123000Z_RRK_FV01_radial.nc': 'IMOS/ACORN/radial_quality_controlled/RRK/2018/11/13',
+    'IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc': 'IMOS/ACORN/radial_quality_controlled/TAN/2019/10/14',
+    'IMOS_ACORN_W_20090808T023000Z_CBG_FV01_wavespec.nc': 'IMOS/ACORN/gridded_1h-avg-wave-spectra_QC/CBG/2009/08/08',
+    'IMOS_ACORN_W_20090801T023000Z_CBG_FV01_wavespec.nc': 'IMOS/ACORN/gridded_1h-avg-wave-spectra_QC/CBG/2009/08/01',
+    'IMOS_ACORN_MW_20110228T073000Z_CBG_FV01_windp.nc': 'IMOS/ACORN/gridded_1h-avg-wind-map_QC/CBG/2011/02/28',
+    'IMOS_ACORN_MW_20110921T063000Z_SAG_FV01_windp.nc': 'IMOS/ACORN/gridded_1h-avg-wind-map_QC/SAG/2011/09/21',
+    'IMOS_ACORN_W_20110921T053000Z_SAG_FV01_wavep.nc': 'IMOS/ACORN/gridded_1h-avg-wave-site-map_QC/SAG/2011/09/21',
+    'IMOS_ACORN_W_20110921T053000Z_CSP_FV01_wavep.nc': 'IMOS/ACORN/gridded_1h-avg-wave-station-map_QC/CSP/2011/09/21',
+    'IMOS_ACORN_V_20110404T003000Z_SAG_FV01_1-hour-avg.nc': 'IMOS/ACORN/gridded_1h-avg-current-map_QC/SAG/2011/04/04',
+    'IMOS_ACORN_V_20150301T013000Z_CBG_FV00_1-hour-avg.nc': 'IMOS/ACORN/gridded_1h-avg-current-map_non-QC/CBG/2015/03/01'
```
ggalibert: Let's remove the FV01 examples above.

test_aodndata/acorn/test_acorn_dm.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestAcornDmHandler(HandlerTestCase):
+    def setUp(self):
+        self.handler_class = AcornHandler
+        super(TestAcornDmHandler, self).setUp()
+
+    def test_various_path(self):
+        for nc_file in ACORN_DM_VARIOUS.keys():
+            dest_path = AcornHandler.dest_path(nc_file)
+            self.assertEqual(dest_path, os.path.join(ACORN_DM_VARIOUS[nc_file], nc_file))
```
ggalibert: Test above is to check destination path is correct but we need to add a test on whether the file is failing regex or not. For example:
```bash
# unknown types
test_unknown_type() {
    local file="IMOS_ACORN_RV_20171212T044500Z_NNB_FV01_unknown.nc"
    local type=`get_type $file`
    assertEquals "type $file" "" "$type"
}

test_match_regex() {
    local good_files bad_files

    good_files="$good_files IMOS_ACORN_RV_20120507T053500Z_CSP_FV01_radial.nc"
    good_files="$good_files IMOS_ACORN_W_20090808T023000Z_CBG_FV01_wavespec.nc"
    good_files="$good_files IMOS_ACORN_MW_20110921T083000Z_SAG_FV01_windp.nc"
    good_files="$good_files IMOS_ACORN_W_20110921T053000Z_SAG_FV01_wavep.nc"
    good_files="$good_files IMOS_ACORN_W_20110921T053000Z_CSP_FV01_wavep.nc"
    good_files="$good_files IMOS_ACORN_V_20150301T013000Z_CBG_FV01_1-hour-avg.nc"

    bad_files="$bad_files IMOS_ACORN_V_20150301T013000Z_CBG_FV00_1-hour-avg.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20171014T060000Z_LANC_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20161121T003000Z_GUI_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_V_20180910T010000Z_BONC_FV00_sea-state.nc"
    bad_files="$bad_files IMOS_ACON_RV_20171014T060000Z_LANC_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_RV_201611203000Z_GUI_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_V_20180910T010000Z_BONC_FV00_sea-se.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20120507T053500Z_FV01_radial.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc.some_suffix"
    bad_files="$bad_files prefix.IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc"

    local file
    for file in $good_files; do
        regex_filter $file
        assertEquals "regex $file" 0 $?
    done

    local file
    for file in $bad_files; do
        regex_filter $file
        assertNotEquals "regex $file" 0 $?
    done
}
```

test_aodndata/acorn/test_acorn_rt.py
```diff
@@ -0,0 +1,56 @@
...
+
+class TestAcornRTHandler(HandlerTestCase):
+    def setUp(self):
+        self.handler_class = AcornHandler
+        super(TestAcornRTHandler, self).setUp()
+
+    def test_various_path(self):
+        for nc_file in ACORN_RT_VARIOUS.keys():
+            dest_path = AcornHandler.dest_path(nc_file)
+            self.assertEqual(dest_path, os.path.join(ACORN_RT_VARIOUS[nc_file], nc_file))
```
ggalibert: Again we need to add a test on whether the file is failing regex or not. For example:
```bash
# unknown types
test_unknown_type() {
    local file="IMOS_ACORN_RV_20171212T044500Z_NNB_FV01_unknown.nc"
    local type=`get_type $file`
    assertEquals "type $file" "" "$type"
}

test_match_regex() {
    local good_files bad_files
    good_files="$good_files IMOS_ACORN_RV_20171014T060000Z_LANC_FV00_radial.nc"
    good_files="$good_files IMOS_ACORN_RV_20161121T003000Z_GUI_FV00_radial.nc"
    good_files="$good_files IMOS_ACORN_V_20180910T010000Z_BONC_FV00_sea-state.nc"
    good_files="$good_files IMOS_ACORN_V_20150301T013000Z_CBG_FV00_1-hour-avg.nc"

    bad_files="$bad_files IMOS_ACORN_RV_20120507T053500Z_CSP_FV01_radial.nc"
    bad_files="$bad_files IMOS_ACORN_W_20090808T023000Z_CBG_FV01_wavespec.nc"
    bad_files="$bad_files IMOS_ACORN_MW_20110921T083000Z_SAG_FV01_windp.nc"
    bad_files="$bad_files IMOS_ACORN_W_20110921T053000Z_SAG_FV01_wavep.nc"
    bad_files="$bad_files IMOS_ACORN_W_20110921T053000Z_CSP_FV01_wavep.nc"
    bad_files="$bad_files IMOS_ACORN_V_20150301T013000Z_CBG_FV01_1-hour-avg.nc"
    bad_files="$bad_files IMOS_ACON_RV_20171014T060000Z_LANC_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_RV_201611203000Z_GUI_FV00_radial.nc"
    bad_files="$bad_files IMOS_ACORN_V_20180910T010000Z_BONC_FV00_sea-se.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20120507T053500Z_FV01_radial.nc"
    bad_files="$bad_files IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc.some_suffix"
    bad_files="$bad_files prefix.IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc"

    local file
    for file in $good_files; do
        regex_filter $file
        assertEquals "regex $file" 0 $?
    done

    local file
    for file in $bad_files; do
        regex_filter $file
        assertNotEquals "regex $file" 0 $?
    done
}
```

test_aodndata/acorn/test_acorn_dm.py
lbesnard: yeap I need to add this

aodndata/acorn/handler.py
```diff
@@ -48,22 +46,55 @@ def __init__(self, *args, **kwargs):
...
+    def get_creation_date(netcdf_path):
+        with Dataset(netcdf_path, mode='r') as nc_obj:
+            nc_pipeline_date_created_str = nc_obj.date_created
+            nc_pipeline_date_created = datetime.strptime(nc_pipeline_date_created_str, '%Y-%m-%dT%H:%M:%SZ')
+        return nc_pipeline_date_created
+
     @staticmethod
     def dest_path(filepath):
-        m = re.search(r'^IMOS_ACORN_([A-Z].*)_([0-9]{8}T[0-9]{6}Z)_([A-Z].*)_(FV0[0-2])_(.*)\.nc$', filepath)
+        m = re.search(r'^IMOS_ACORN_([A-Z].*)_([0-9]{8}T[0-9]{6}Z)_([A-Z]{3,4})_(FV0[0-1])_(.*)\.nc$',
```
ggalibert: Just like in the original regex:
```bash
ACORN_RT_REGEX='^IMOS_ACORN_[[:alpha:]]{1,2}_[[:digit:]]{8}T[[:digit:]]{6}Z_[[:alpha:]]{3,4}_FV00_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$'
```
Can we please continue to restrict on the last bit on the file name before the extension? -> (radial|sea-state|wavespec|windp|wavep|1-hour-avg)

aodndata/acorn/handler.py
```diff
@@ -48,22 +46,55 @@ def __init__(self, *args, **kwargs):
...
+
+        # if file contains any of FILE_TYPE_NEED_INDEX, we index, otherwise publish only
+        if any(s in get_type(nc_collection._name) for s in FILE_TYPE_NEED_INDEX):
+            nc_collection.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
+        else:
+            nc_collection.publish_type = PipelineFilePublishType.UPLOAD_ONLY
+
+        # check if file with same dest path already on s3. If yes, check it's date_created nc attribute to know
+        # if we need to overwrite this object or not
+            destination_s3 = self.dest_path(nc_collection._name)
```
ggalibert: Indent?

test_aodndata/acorn/test_acorn_rt.py
```diff
@@ -24,22 +22,6 @@
...
-    'IMOS_ACORN_RV_20181113T123000Z_RRK_FV01_radial.nc': 'IMOS/ACORN/radial_quality_controlled/RRK/2018/11/13',
-    'IMOS_ACORN_RV_20191014T192000Z_TAN_FV01_radial.nc': 'IMOS/ACORN/radial_quality_controlled/TAN/2019/10/14',
-    'IMOS_ACORN_W_20090808T023000Z_CBG_FV01_wavespec.nc': 'IMOS/ACORN/gridded_1h-avg-wave-spectra_QC/CBG/2009/08/08',
-    'IMOS_ACORN_W_20090801T023000Z_CBG_FV01_wavespec.nc': 'IMOS/ACORN/gridded_1h-avg-wave-spectra_QC/CBG/2009/08/01',
-    'IMOS_ACORN_MW_20110228T073000Z_CBG_FV01_windp.nc': 'IMOS/ACORN/gridded_1h-avg-wind-map_QC/CBG/2011/02/28',
-    'IMOS_ACORN_MW_20110921T063000Z_SAG_FV01_windp.nc': 'IMOS/ACORN/gridded_1h-avg-wind-map_QC/SAG/2011/09/21',
-    'IMOS_ACORN_W_20110921T053000Z_SAG_FV01_wavep.nc': 'IMOS/ACORN/gridded_1h-avg-wave-site-map_QC/SAG/2011/09/21',
-    'IMOS_ACORN_W_20110921T053000Z_CSP_FV01_wavep.nc': 'IMOS/ACORN/gridded_1h-avg-wave-station-map_QC/CSP/2011/09/21',
-    'IMOS_ACORN_V_20110404T003000Z_SAG_FV01_1-hour-avg.nc': 'IMOS/ACORN/gridded_1h-avg-current-map_QC/SAG/2011/04/04',
-    'IMOS_ACORN_V_20150301T013000Z_CBG_FV00_1-hour-avg.nc': 'IMOS/ACORN/gridded_1h-avg-current-map_non-QC/CBG/2015/03/01'
```
ggalibert: Oops, you should keep that last one.

lbesnard:
@ggalibert added some code to make sure FV00 and FV01 don't land in wrong folders. This is then enabled (but tested here in the unittests) with proper configuration in the json files on chef-private

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,106 @@
...
+    @staticmethod
+    def get_creation_date(netcdf_path):
+        with Dataset(netcdf_path, mode='r') as nc_obj:
+            nc_pipeline_date_created_str = nc_obj.date_created
+            nc_pipeline_date_created = datetime.strptime(nc_pipeline_date_created_str, '%Y-%m-%dT%H:%M:%SZ')
+        return nc_pipeline_date_created
+
+    @staticmethod
+    def dest_path(filepath):
+        m = re.search(r'^IMOS_ACORN_([A-Z].*)_([0-9]{8}T[0-9]{6}Z)_([A-Z]{3,4})_(FV0[0-1])_(.*)\.nc$',
```
lwgordonimos: I think named groups are really good for when there is a strict naming convention like this. It makes the code later on much more readable than having to later map a numeric index to one of the fields in the file name.

For example, changing nothing else other than adding the named fields:

```python
    @staticmethod
    def dest_path(filepath):
        pattern = re.compile(r"""
                        ^IMOS_ACORN_
                        (?P<named_field1>[A-Z].*)_
                        (?P<nc_time_cov_start>[0-9]{8}T[0-9]{6}Z)_
                        (?P<platform_code>[A-Z]{3,4})_
                        (?P<another_field>FV0[0-1])_
                        (?P<yet_another_field>.*)
                        \.nc$
                        """, re.VERBOSE)

        m = pattern.match(os.path.basename(filepath))
        if m is None:
            raise InvalidFileNameError(
                "file name: \"{filename}\" not matching regex to deduce dest_path for ACORN".format(
                    filename=os.path.basename(filepath)))

        fields = m.groupdict()

        nc_time_cov_start = datetime.strptime(fields['nc_time_cov_start'], '%Y%m%dT%H%M%SZ')

        file_type = get_type(filepath)
        return os.path.join('IMOS', 'ACORN', file_type, fields['platform_code'],
                            '%d' % nc_time_cov_start.year, '%02d' % nc_time_cov_start.month,
                            '%02d' % nc_time_cov_start.day,
                            os.path.basename(filepath)
                            )
```

lbesnard: thanks I didn't know how to do that

ggalibert: @lbesnard can you please update your code following @lwgordonimos 's recommendations above?
Fields are data_codes, time_coverage_start, platform_code, file_version and file_type_code.

test_aodndata/acorn/test_acorn.py
```diff
@@ -0,0 +1,84 @@
...
+
+from aodncore.pipeline import PipelineFilePublishType
+from aodncore.pipeline.exceptions import InvalidFileNameError
+from aodndata.acorn.handler import AcornHandler
+from aodncore.testlib import HandlerTestCase
+
+TEST_ROOT = os.path.join(os.path.dirname(__file__))
+
+GOOD_NC_FV00 = os.path.join(TEST_ROOT,
+                       'IMOS_ACORN_V_20180501T003000Z_ROT_FV00_1-hour-avg.nc')
```
ggalibert: Why do we need to add some actual files on GitHub? Aren't the tests only performed on file names?

lwgordonimos: For tests that only deal with names (strings) this is true, but some real files are useful for other tests like running a file through the whole handler, or for things that need to extract variables etc.

In some cases it might be appropriate to mock out the bits that deal with actual files, but sometimes it's really handy to have a file which you *know* should always pass through with an expected result and a file which you know should fail (and if not, the lack of a failure is considered an error).

For example GOOD_NC might be a file you want to say "this file should always be handled successfully, regardless of how the *code* is changed" and similar for a "this file should always be rejected because of a check or because it was missing some variable that the handler needed to generate the dest path".

aodndata/acorn/handler.py
lbesnard: yeap already done, but haven't pushed it yet

test_aodndata/acorn/test_acorn.py
lbesnard: The content of the file is also tested here, to retrieve the date_create attribute

test_aodndata/acorn/test_acorn.py
```diff
@@ -0,0 +1,84 @@
...
+                                                   os.path.basename(GOOD_NC_FV00)))
+        self.assertEqual(f.publish_type, PipelineFilePublishType.HARVEST_UPLOAD)
+        self.assertTrue(f.is_checked)
+        self.assertTrue(f.is_stored)
+        
+    def test_fv00_in_dm(self):
+        # check that a FV00 file is not allowed in DM folder/pipeline
+        self.run_handler_with_exception(InvalidFileNameError, GOOD_NC_FV00,
+                                        include_regexes=[
+                                            '^IMOS_ACORN_[A-Z].*_[0-9]{8}T[0-9]{6}Z_[A-Z]{3,4}_FV01_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$']
```
ggalibert: If I understand well, this regex will be set somewhere else in the pipeline. It would be good to have single entry point definition for this regex that could be re-used here.

lbesnard: The regexp is set up here 
https://github.com/aodn/chef-private/pull/2842/files

lbesnard: the one in the test_acorn is only there as a test, to force a config file in a way.

ggalibert: Can you retrieve the regex from the .json rather than duplicating it in your unit test? Just in case we update the regex in the future...

lbesnard: not really. The unittest have to be successful independently of the config.

test_aodndata/acorn/test_acorn.py
ggalibert: The test on creation_date didn't exist in previous pipeline though. I'm hoping we're not going too far in what is just a migration.

lbesnard: well you can just test about anything in unittests. It just testing that the function is behaving fine, which means that in case someone changes this function, well the unittest will fail accordingly. It makes it easier to program something knowing you 're not going to break things expectingly.

It is a migration, but with a total different design in mind

test_aodndata/acorn/test_acorn.py
```diff
@@ -0,0 +1,84 @@
...
+                                   include_regexes=['IMOS_ACORN_.*\.nc']
+                                   )
+        self.assertEqual(len(handler.file_collection), 1)
+        f = handler.file_collection[0]
+
+        self.assertEqual(f.dest_path, os.path.join('IMOS/ACORN/gridded_1h-avg-current-map_non-QC/ROT/2018/05/01',
+                                                   os.path.basename(GOOD_NC_FV00)))
+        self.assertEqual(f.publish_type, PipelineFilePublishType.HARVEST_UPLOAD)
+        self.assertTrue(f.is_checked)
+        self.assertTrue(f.is_stored)
```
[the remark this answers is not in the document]
lbesnard: nop we don't push files to s3 or compare to existing files on s3. the unittests are a different layer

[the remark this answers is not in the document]
lbesnard: we can however test the functionality by faking it. not testing here yet since the code to check another file on s3 is not in place yet

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,115 @@
...
+        #storage_query_res = self.state_query.query_storage(destination_s3)
+        #if storage_query_res:
+        #    creation_date_nc_pipeline = self.get_creation_date(nc_collection)  # creation date of the file in the pipeline
+
+        #    #TODO get leigh to create the funcion below
+        #    nc_s3_path = download_s3(destination_s3)
+        #    creation_date_nc_s3 = get_creation_date(nc_s3_path)  # creation date of the file on s3
+        #
+        #    if creation_date_nc_pipeline > creation_date_nc_s3:
+        #        nc_collection.publish_type = PipelineFilePublishType.NO_ACTION
```
lbesnard: Not to forget this section still needs to be written. Just waiting for @lwgordonimos to think about this functionality

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,115 @@
...
+    @staticmethod
+    def get_creation_date(netcdf_path):
+        with Dataset(netcdf_path, mode='r') as nc_obj:
+            nc_pipeline_date_created_str = nc_obj.date_created
+            nc_pipeline_date_created = datetime.strptime(nc_pipeline_date_created_str, '%Y-%m-%dT%H:%M:%SZ')
+        return nc_pipeline_date_created
+
+    @staticmethod
+    def dest_path(filepath):
+        pattern = re.compile(r"""
```
lwgordonimos: Now that I look at it, maybe this could be split into a "get_fields_from_filename" function which takes the filename as a parameter and returns the dict.

Also move the pattern to a module level global like ACORN_FILE_PATTERN. Would keep the dest_path method more lean and mean, and improves testability.

aodndata/acorn/handler.py
```diff
@@ -34,31 +34,49 @@ def get_fields_from_filename(filename):
...
-        if any(s in filepath for s in site_map_station):
-            acorn_file_type = "gridded_1h-avg-wave-site-map_QC"
         else:
-            acorn_file_type = "gridded_1h-avg-wave-station-map_QC"
-    elif "FV00_1-hour-avg.nc" in filepath:
-        acorn_file_type = "gridded_1h-avg-current-map_non-QC"
-    elif "FV01_1-hour-avg.nc" in filepath:
-        acorn_file_type = "gridded_1h-avg-current-map_QC"
+            raise InvalidFileNameError("file name: \"{filename}\" Unknown product type from filename".
+                                       format(filename=os.path.basename(filepath)))
```
ggalibert: Could have this line present only once in your function (instead of being repeated twice). Error would be raised if acorn_file_type hasn't been set for example.

lbesnard: definitely

lbesnard:
@ggalibert  done

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,134 @@
...
+        raise InvalidFileNameError("file name: \"{filename}\" Unknown product type from filename".
+                                   format(filename=os.path.basename(filepath)))
+
+    return acorn_file_type
+
+
+class AcornHandler(HandlerBase):
+    def __init__(self, *args, **kwargs):
+        super(AcornHandler, self).__init__(*args, **kwargs)
+        self.allowed_extensions = ['.nc']
```
lwgordonimos:
```python
# handler parameters you would add:
custom_params={'thredds_url': 'http://thredds.aodn.org.au/thredds/dodsC/'}

#then *inside* the handler class, you can access it like this:
self.custom_params.get('thredds_url', 'http://thredds.aodn.org.au/thredds/dodsC/')
```

aodndata/acorn/handler.py
```diff
@@ -79,6 +79,7 @@ class AcornHandler(HandlerBase):
     def __init__(self, *args, **kwargs):
         super(AcornHandler, self).__init__(*args, **kwargs)
         self.allowed_extensions = ['.nc']
+        self.opendap_root = self.resolve_params['opendap_root']
```
lwgordonimos: If it's in the config, it will not be in the params, so you can access it like so:

```python
self.opendap_root = self.config.pipeline_config['global'].get('opendap_root', 'http://thredds.aodn.org.au/thredds/dodsC/')
```

lbesnard:
@lwgordonimos just did the change to use opendap_root properly. So maybe we can deploy this this week on RC and do some final testing with the help of @ggalibert ?

lwgordonimos:
@lbesnard there was a fundamental and irreconcilable logic problem preventing us from going forward with one of the pipelines without core changes, but I think that was GSLA with the deletes? If so, we should be able to move forward with this after looking at the PR again as a whole.

lbesnard:
agree, will prob squash it all

lbesnard:
@lwgordonimos do you think we could test this on RC today ?

lbesnard:
https://github.com/aodn/chef-private/pull/2948

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+from aodncore.pipeline import HandlerBase, PipelineFilePublishType
+from aodncore.pipeline.exceptions import InvalidFileNameError
+from aodncore.util.misc import get_pattern_subgroups_from_string
+
+FILE_TYPE_NEED_INDEX = ('radial', 'radial_quality_controlled', 'gridded_1h-avg-current-map_non-QC',
+                        'gridded_1h-avg-current-map_QC', 'gridded_1h-avg-wind-map_QC',
+                        'gridded_1h-avg-wave-site-map_QC')
+
+ACORN_FILE_PATTERN = re.compile(r"""
+                                (.*/|)IMOS_ACORN_
```
lwgordonimos: What's the reason for the extra wildcard stuff on the front of the regex? Would be cleaner if this was just the filename and things using the regex were expected to use the basename, because the directory shouldn't ever really be considered when matching

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+            acorn_file_type = "gridded_1h-avg-current-map_QC"
+
+        else:
+            unknown_product = True
+    else:
+        unknown_product = True
+
+    if unknown_product:
+        raise InvalidFileNameError("file name: \"{filename}\" Unknown product type from filename".
+                                   format(filename=os.path.basename(filepath)))
```
lwgordonimos: You've already got the basename in a variable (file_basename), so can avoid this extra function call:

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+        super(AcornHandler, self).__init__(*args, **kwargs)
+        self.allowed_extensions = ['.nc']
+        self.opendap_root = self.config.pipeline_config['global'].get('opendap_root')
+
+    def preprocess(self):
+        nc_collection = self.file_collection[0]
+
+        # enforce a check on the include_regexes and the filename to make sure the handler is processing the correct
+        # matching rexep file written in the config file
+        m = re.search(self.include_regexes[0], nc_collection.name)
```
lwgordonimos: This should be done using core functionality, by using the "allowed_regexes" feature:

https://aodn.github.io/python-aodncore/aodncore.pipeline.html?highlight=handlerbase#aodncore.pipeline.HandlerBase

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class AcornHandler(HandlerBase):
+    def __init__(self, *args, **kwargs):
+        super(AcornHandler, self).__init__(*args, **kwargs)
+        self.allowed_extensions = ['.nc']
+        self.opendap_root = self.config.pipeline_config['global'].get('opendap_root')
+
+    def preprocess(self):
+        nc_collection = self.file_collection[0]
```
lwgordonimos: This name is confusing, since it is referring to a single file but the variable is "collection". Rename to "nc_file"?

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+            nc_collection.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
+        else:
+            nc_collection.publish_type = PipelineFilePublishType.UPLOAD_ONLY
+
+        # check if file with same dest path already on s3. If yes, check its date_created nc attribute to know
+        # if we need to overwrite this object or not
+        destination_s3 = self.dest_path(nc_collection.name)
+
+        storage_query_res = self.state_query.query_storage(destination_s3)
+        if storage_query_res:
```
lwgordonimos: This could be made more explicit to help avoid those weird S3 bugs we were debugging with additional extensions:

```python
if destination_s3 in storage_query_res:
```

This checks whether the *exact* name is in the storage already, not just the *prefix*

aodndata/acorn/handler.py
```diff
@@ -0,0 +1,133 @@
...
+            creation_date_nc_pipeline = self.get_creation_date(nc_collection.src_path)
+
+            # creation date of the file already published
+            creation_date_nc_s3 = self.get_creation_date(os.path.join(self.opendap_root, destination_s3))
+
+            if creation_date_nc_pipeline < creation_date_nc_s3:
+                nc_collection.publish_type = PipelineFilePublishType.NO_ACTION
+
+    @staticmethod
+    def get_creation_date(netcdf_path):
```
lwgordonimos: This should be moved out of the class into a separate function, since it is static and not related directly to the class.

test_aodndata/acorn/test_acorn_dm.py
ggalibert: any progress on this?

test_aodndata/acorn/test_acorn_rt.py
ggalibert: Progress?

aodndata/acorn/handler.py
```diff
@@ -68,55 +68,49 @@ def get_type(filepath):
...
         self.allowed_extensions = ['.nc']
         self.opendap_root = self.config.pipeline_config['global'].get('opendap_root')
+        self.allowed_regexes = [ACORN_FILE_PATTERN]
 
     def preprocess(self):
-        nc_collection = self.file_collection[0]
-
-        # enforce a check on the include_regexes and the filename to make sure the handler is processing the correct
-        # matching rexep file written in the config file
-        m = re.search(self.include_regexes[0], nc_collection.name)
```
ggalibert: Why is this test not performed any more?

test_aodndata/acorn/test_acorn.py
```diff
@@ -51,16 +51,18 @@ def test_good_netcdf_generic_fv00(self):
         
     def test_fv00_in_dm(self):
         # check that a FV00 file is not allowed in DM folder/pipeline
-        self.run_handler_with_exception(InvalidFileNameError, GOOD_NC_FV00,
+        self.run_handler_with_exception(AttributeValidationError, GOOD_NC_FV00,
```
ggalibert: Why this change? Aren't we checking the file name?

test_aodndata/acorn/test_acorn.py
```diff
@@ -51,16 +51,18 @@ def test_good_netcdf_generic_fv00(self):
...
                                         include_regexes=[
-                                            '^IMOS_ACORN_[A-Z].*_[0-9]{8}T[0-9]{6}Z_[A-Z]{3,4}_FV01_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$']
+                                            '^IMOS_ACORN_[A-Z].*_[0-9]{8}T[0-9]{6}Z_[A-Z]{3,4}_FV01_(radial|sea-state|wavespec|windp|wavep|1-hour-avg)\.nc$'],
+                                            allowed_dest_path_regexes=["^IMOS/ACORN/.*FV01.*\.nc$"]
```
ggalibert: Fix indent?

test_aodndata/acorn/test_acorn_dm.py
lbesnard: yeap 
in test_fv00_in_dm and test_fv01_in_rt

test_aodndata/acorn/test_acorn_rt.py
lbesnard: same as above

PR #47, opened 2018-06-04
Suggested enhancements to the 'get_previous_version' method

lwgordonimos:
I had some difficulty working out the logic going on in this particular method, so I just quickly added what I interpreted it to be doing in the way I would probably tackle it to avoid the nesting that makes it harder to read and test.


bpasquer:
the suggested enhancements have been merged into the soop_ba branch. closing the PR

PR #50, opened 2018-06-14
MooringsHandler: code clean up and update

mhidas:
This PR doesn't change the behaviour of the MooringsHandler, its just to "modernise" the code to make use of new helper methods that were introduced after the handler was created. The ABOS handler will be a subclass of this one, so I'd like to get this in before that one.

### DIFF
--- a/aodndata/moorings/handlers.py
+++ b/aodndata/moorings/handlers.py
@@ -1,4 +1,4 @@
-from aodncore.pipeline import HandlerBase, PipelineFilePublishType, PipelineFileCheckType
+from aodncore.pipeline import HandlerBase, PipelineFilePublishType, PipelineFileCheckType, FileType
 from aodncore.pipeline.exceptions import InvalidFileNameError
 from aodndata.moorings.classifiers import MooringsFileClassifier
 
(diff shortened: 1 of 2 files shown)


aodndata/moorings/handlers.py
```diff
@@ -38,19 +31,16 @@ def preprocess(self):
...
 
-        invalid_files = []
-        for f in self.file_collection:
-            if f.publish_type == PipelineFilePublishType.NO_ACTION:
-                invalid_files.append(f.name)
-            if f.extension != '.nc':
-                f.check_type = PipelineFileCheckType.NO_ACTION
-                f.publish_type = PipelineFilePublishType.UPLOAD_ONLY
-
+        invalid_files = self.file_collection.filter_by_attribute_id('publish_type', PipelineFilePublishType.NO_ACTION)
```
lwgordonimos: This check would be unnecessary when this is merged, so wonder if we should aim to just merge that?

https://github.com/aodn/python-aodncore/pull/96

mhidas: You're probably right, but I think you mean https://github.com/aodn/python-aodncore/pull/91 no? (he new default publish type)
That's already merged, and I've just built it, so I'll take a look and update this PR as necessary.

mhidas: Also, I still prefer to do this here in the pre-process step, and throw a handler-specific `InvalidFileNameError` that can be sent to the uploader. Otherwise we keep going to the publish step, and end up with an `UnmatchedFilesError` which is harder to debug and is only sent to me.

I just need to update `NO_ACTION` to `UNSET`.

lwgordonimos: Makes sense. I think you're right in that there are two different concepts in play.

The 'UnmatchedFilesError' implies that the *handler* had the problem by allowing it to get that far, so notifying the owner is appropriate. The preprocess check is a different animal, so I think it's appropriate to handle it as you want in a per-pipeline manner.

PR #51, opened 2018-06-14
Update GenericHandler and MooringsHandler to work with aodncore 0.12

mhidas:
This is to account for the new default publish_type introduced by https://github.com/aodn/python-aodncore/pull/91


mhidas:
Travis checks will fail because aodncore 0.12 is not in prod yet

mhidas:
... but we can't deploy aodncore 0.12 to prod before these changes because it will break these handlers! :confounded: Humph!

PR #52, opened 2018-06-14
SRS - use get_pattern_subgroups_from_string

lbesnard:


aodndata/srs/srs_ghrsst.py
```diff
@@ -2,19 +2,21 @@
...
 from datetime import datetime
 
+from aodncore.util.misc import get_pattern_subgroups_from_string
+
 from aodncore.pipeline import HandlerBase
 from aodncore.pipeline.exceptions import InvalidFileNameError
 
 L3S_L3C_FILE_PATTERN = re.compile(r"""
-                                ^(?P<nc_time_cov_start>[0-9]{14})-ABOM-
+                                (.*/|)(?P<nc_time_cov_start>[0-9]{14})-ABOM-
```
lwgordonimos: What was the reason for this change to the pattern? Looks a bit odd to me...

lwgordonimos: https://github.com/aodn/python-aodndata/pull/52/files#diff-62a887b1575ecffb84e3cff79877c85eR42

But you are already grabbing the basename here for this purpose?

lbesnard: good point, but I should change it here 
https://github.com/aodn/python-aodndata/pull/52/files#diff-62a887b1575ecffb84e3cff79877c85eR45

lbesnard:
@lwgordonimos rectified

aodndata/srs/srs_oc_gridded.py
```diff
@@ -14,32 +15,21 @@
 OC_GRIDDED_PREFIX_PATH = 'IMOS/SRS/OC/gridded'
 
 RJOHNSON_FILE_PATTERN = re.compile(r"""
-                                ^(?P<data_parameter_code>A|S)
+                                (?P<data_parameter_code>A|S)
```
lwgordonimos: Not sure what the best practice is with the carets when using re.match... they seem redundant, but the pattern is separate from the match function, so I wonder if we should keep them in?

PR #54, opened 2018-06-19
RAMSSA/GAMSSA rename test file in unittest. Does not break handler

lbesnard:


lbesnard:
* upadte unittest to match new filename given by facility

PR #66, opened 2018-08-09
add processing of ANFOG DM .txt and DSTG .rar archive file

bpasquer:


bpasquer:
The PR is only changing 2 regex. Please merge and deploy

PR #70, opened 2018-08-14
ACORN - add logging INFO/WARNING to handler

lbesnard:


lbesnard:
@lwgordonimos suggested we could use sumologic

aodndata/acorn/handler.py
```diff
@@ -109,7 +109,12 @@ def preprocess(self):
             creation_date_nc_s3 = get_creation_date(os.path.join(self.opendap_root, destination_s3))
 
             if creation_date_nc_pipeline < creation_date_nc_s3:
+                self.logger.warning("incoming file '{dest_path}': creation date is older than file already on storage".
```
ggalibert: Can we add in the message something like "publish_type set to NO_ACTION"?

aodndata/acorn/handler.py
```diff
@@ -109,7 +109,12 @@ def preprocess(self):
...
+                                 format(dest_path=nc_file.name))
                 nc_file.publish_type = PipelineFilePublishType.NO_ACTION
+            else:
+                self.logger.info("incoming file '{dest_path}': creation date is newer/equal to file already on storage".
+                                 format(dest_path=nc_file.name))
```
ggalibert: This is the expected use case, I don't think it's worth logging this, this would make the logs more noisy than anything.

aodndata/acorn/handler.py
lwgordonimos: That kind of message is already logged elsewhere in an INFO message, since there is a callback which prints every file status change automatically, for example:

```
2018-08-14 12:49:32,353 INFO [root] updated file 'IMOS_ACORN_V_20170101T003000Z_ROT_FV01_1-hour-avg.nc': {'publish_type': 'NO_ACTION'}
```

So the warning message is really to complement this standard state change message.

ggalibert: Ok thanks, didn't see this in the logs.

lbesnard:
@ggalibert done

PR #76, opened 2018-08-15
GSLA - move file with older creation date to ERROR_DIR

lbesnard:


aodndata/gsla/handler.py
```diff
@@ -3,7 +3,7 @@
 from datetime import datetime
 
 from aodncore.pipeline import HandlerBase, FileType, PipelineFilePublishType, PipelineFile
-from aodncore.pipeline.exceptions import InvalidFileNameError, InvalidInputFileError, InvalidPathFunctionError
```
ggalibert: Is this change intended?

lbesnard: yes, because it wasn't used anyway
https://github.com/aodn/python-aodndata/blob/master/aodndata/gsla/handler.py

PR #81, opened 2018-09-06
Add setting of error_cleanup_regexes to MooringsHandler

mhidas:


mhidas:
Travis failing, apparently due to errors building cf-units 2.0.2 (which is now required by compliance-checker 4.1.1 - see https://github.com/ioos/compliance-checker/pull/588).

mhidas:
Rebased onto master, whichshould fix Travis

mhidas:
I've rebased this again, and updated the unittests so that they test the value of `error_cleanup_regexes` as well. Should be good to go now.

PR #82, opened 2018-09-07
ACORN - add postprocess to rm file from error_dir

lbesnard:


aodndata/acorn/handler.py
```diff
@@ -112,6 +112,14 @@ def preprocess(self):
                 raise InvalidFileContentError("file name: \"{filename}\"  creation date is older than file already on storage".
                                            format(filename=nc_file.name))
 
+    def postprocess(self):
+        """Set error_cleanup_regexes so that if the same file was uploaded previously and failed, it can now be
+        removed from the error directory (by setting "success_exit_policies" in the pipeline config to
+        "DELETE_CUSTOM_REGEXES_FROM_ERROR_STORE").
+        """
+        self.error_cleanup_regexes = self.file_basename
```
mhidas: If you just want to delete error files with the same exact basename, you don't need to set this attribute in the handler. Instead, use the exit policy "DELETE_SAME_NAME_FROM_ERROR_STORE" (see https://github.com/aodn/python-aodncore/pull/117#issue-209063959)

lbesnard:
Thanks @mhidas 
https://github.com/aodn/chef-private/pull/2989

PR #93, opened 2018-11-28
update AbosHandler to handle images in a more flexible way

mhidas:


mhidas:
This is failing because the required version of aodncore is not in prod yet.
Something like #95 might help with this, though the overall Travis build will still fail

mhidas:
#95 merged, so I've rebased this on top. The `aodncore_build` job [here](https://travis-ci.org/aodn/python-aodndata/builds/464129412) should now pass, meaning this will work as long as it's deployed along with aodncore v0.24.1

mhidas:
@lbesnard @bpasquer @ggalibert Could someone please do a quick review and merge this for me?
It would allow me to clear one of the red rows in Nagios next week.

PR #95, opened 2018-12-05
Use Travis build matrix to test against different versions of aodncore

mhidas:
This is an attempt to deal with cases where we're starting to use new aodncore functionality that hasn't been deployed to prod yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,16 @@ env:
   global:
     - MPLBACKEND=agg
     - AODNFETCHER_URL="git+https://github.com/aodn/python-aodnfetcher.git@master"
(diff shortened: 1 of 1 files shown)


mhidas:
For the latest version of aodncore (0.24.0) we're getting a lot of errors like this:
`AttributeError: 'SoopBaHandler' object has no attribute '_state_query'`

mhidas:
The other issue is that some or all of the artifacts from aodncore_build, aodncore_systest and aodncore_prod might actually be the same. Ideally we'd like to test against all unique versions of aodncore between what's in prod and the latest _build.

lwgordonimos:
Nice! This picked up a bug already.

I think we should be going out of our way *not* to break aodndata, and if something fundamental changes, try to deprecate it over a couple of versions.

PR #98, opened 2018-12-13
SRS SURFACE WAVES - mod handler to deal with manifest_dir

lbesnard:
https://github.com/aodn/chef-private/pull/3107


setup.py
```diff
@@ -28,8 +29,8 @@
         'SrsOcGriddedHandler = aodndata.srs.srs_oc_gridded:SrsOcGriddedHandler',
         'SrsOcLjcoHandler = aodndata.srs.srs_oc_ljco:SrsOcLjcoHandler',
         'SrsOcSoopRadHandler = aodndata.srs.srs_oc_soop_rad:SrsOcSoopRadHandler',
-        'SstaarsHandler = aodndata.csiro.sstaars:SstaarsHandler',
-        'SoopCo2Handler = aodndata.soop.soop_co2:SoopCo2Handler'
```
ggalibert: You removed SoopCo2Handler here, is this normal?

lbesnard: no I moved it above to be ordered properly

aodndata/srs/srs_surface_waves.py
```diff
@@ -20,25 +21,30 @@
...
-            "file name: \"{filename}\" not matching regex to deduce dest_path".format(
-                filename=os.path.basename(filepath)))
-
-    return os.path.join(PREFIX_PATH, fields['platform_code'], file_basename)
-
-
+class SrsSurfaceWavesHandler(HandlerBase):
+    def __init__(self, *args, **kwargs):
+        super(SrsSurfaceWavesHandler, self).__init__(*args, **kwargs)
+        self.allowed_extensions = ['.nc', '.dir_manifest']
```
ggalibert: I think what we'll want is to allow manifest files only.
The pipeline should watch a 'manifest' directory rather than the current incoming directory. The incoming directory stays for the user to be able to upload its data files but they don't trigger anything. Then when we upload a manifest in the 'manifest' directory it triggers the pipeline and the manifest points to the files in incoming.

lbesnard:
after chatting with @lwgordonimos  we disregard this PR

PR #99, opened 2018-12-17
SRS SURFACE WAVES - destpath create dir per bin

lbesnard:


aodndata/srs/srs_surface_waves.py
```diff
@@ -26,19 +26,45 @@ def dest_path_srs_surface_waves(filepath):
...
                                 """.format(platforms), re.VERBOSE)
 
     file_basename = os.path.basename(filepath)
     if FILE_PATTERN.match(file_basename):
         fields = get_pattern_subgroups_from_string(file_basename, FILE_PATTERN)
+        lat = int(fields['latitude'])
+        lon = int(fields['longitude'])
+        lat_dir = fields['latitude_dir']
+
+        def coord_lower_limit_val(val):
```
ggalibert: You can rename the function `10_rounding_to_floor` ?

aodndata/srs/srs_surface_waves.py
```diff
@@ -26,19 +26,45 @@ def dest_path_srs_surface_waves(filepath):
...
+            """ return the bin bottom left coordinate of a coordinate value
+            exampe: 26 -> 20
+                    31 -> 30
+                    0 -> 0
+                    10 -> 10
+                    """
+            return int(round(val / 10) * 10)
```
ggalibert: return int(val / 10) * 10

lwgordonimos: This provides a more concise method of achieving this without any floating point rounding or type coercion required:

https://stackoverflow.com/a/13082705

e.g.

```python
def round_down(num, divisor=10):
    return num - (num % divisor)
```

PR #101, opened 2018-12-19
Pipeline V2 processing for MHL files

bpasquer:


test_aodndata/aodn_wave_dm/test_aodn_wave_dm.py
```diff
@@ -74,6 +84,5 @@ def test_dest_path_aodn_wave_dm(self):
         with self.assertRaises(InvalidFileNameError):
             dest_path_aodn_wave_dm(bad)
 
-
-if __name__ == '__main__':
-    unittest.main()
+        if __name__ == '__main__':
+            unittest.main()
```
mhidas: These last two lines should not be indented

bpasquer: well spotted . I'l put up a PR to fix it https://github.com/aodn/python-aodndata/pull/102

PR #103, opened 2018-12-20
SOOP XBT DM - memory leak fix attempt

lbesnard:
@lwgordonimos don't know if that would be enough, or if this was the cause of the memory leak issue. But this works

### DIFF
--- a/aodndata/soop/soop_xbt_dm.py
+++ b/aodndata/soop/soop_xbt_dm.py
@@ -1,5 +1,4 @@
 import datetime
-import itertools
 import os
 import tempfile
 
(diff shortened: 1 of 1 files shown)


aodndata/soop/soop_xbt_dm.py
```diff
@@ -42,42 +41,52 @@ def dest_path(filepath):
...
+        temp_qc = F.variables['TEMP_quality_control'][:]
+        time_coverage_start = F.time_coverage_start
+        temp_values = sea_water_temperature[:]
+        depth_values = depth[:]
+        sst_fillvalue = sea_water_temperature._FillValue
+        sst_units = sea_water_temperature.units
+        sst_longname = sea_water_temperature.long_name
+        depth_units = depth.units
+        depth_longname = depth.long_name
+        if hasattr(depth, '_FillValue'):
```
lwgordonimos: I'd do it like this, since querying the scope via locals() is a bit clunky:

```python
try:
    depth_fillvalue = depth._FillValue
except AttributeError:
    depth_fillvalue = None
```

aodndata/soop/soop_xbt_dm.py
```diff
@@ -42,42 +41,52 @@ def dest_path(filepath):
...
-    temp_values[~i_good_data] = sea_water_temperature._FillValue
+    temp_values[~i_good_data] = sst_fillvalue
 
     # new XBT files dont have a FillValue att for DEPTH since DEPTH is a
     # dimension. However previous files do. Need to handle both case if we do
     # some reprocessing
-    if hasattr(depth, '_FillValue'):
-        depth_values[~i_good_data] = depth._FillValue
-        depth_values = ma.masked_values(depth_values, F.variables['DEPTH']._FillValue)
+    if 'depth_fillvalue' in locals():
```
lwgordonimos: With the above change, this then becomes:

```python
if depth_fillvalue:
   # do stuff
```

aodndata/soop/soop_xbt_dm.py
lwgordonimos: Even better

lbesnard:
@lwgordonimos  done !

lwgordonimos:
@lbesnard it looks ok to me, so happy to merge it and move to testing on RC

PR #106, opened 2019-01-08
SOOP TRV - rm duplicates via WFS - WIP

lbesnard:


lbesnard:
@lwgordonimos thanks a lot. I'm not quite sure I understand your changes, but I added mine on top. Looks like it's doing what I want.

PR #107, opened 2019-01-10
SRS GHRSST - add path creation for VIIRS and multisensor data

lbesnard:


aodndata/srs/srs_ghrsst.py
```diff
@@ -10,20 +10,45 @@
...
                                 (?P<nc_time_cov_start>[0-9]{14})-ABOM-
                                 (?P<product_type>L3U)_.*-AVHRR
-                                (?P<sat_number>[0-9].*)_D-
+                                (?P<sat_value>[0-9].*)_D-
                                 (?P<pass_direction>Asc|Des)
                                 (?P<optional>_Southern|)
                                 \.nc$
                                 """, re.VERBOSE)
 
+L3U_VIRRS_FILE_PATTERN = re.compile(r"""
```
bpasquer: There is a difference in the file pattern : is it VIRRS or VIIRS?

aodndata/srs/srs_ghrsst.py
```diff
@@ -10,20 +10,45 @@
...
+                                (?P<nc_time_cov_start>[0-9]{14})-ABOM-
+                                (?P<product_type>L3U)_.*-VIIRS_NPP.*
+                                (?P<optional>_Southern|)
+                                \.nc$
+                                """, re.VERBOSE)
+
+L3C_VIRRS_FILE_PATTERN = re.compile(r"""
```
bpasquer: same as above

aodndata/srs/srs_ghrsst.py
lbesnard: thanks. will correct. it's viirs

lbesnard: @bpasquer done !

PR #110, opened 2019-01-31
Update product codes

bpasquer:
Add code related to transect to Sydney


bpasquer:
@lbesnard  change is required to process  some files in the error dir

PR #117, opened 2019-03-04
Soop tmv

lbesnard:


aodndata/soop/soop_tmv_nrt.py
```diff
@@ -134,28 +174,44 @@ def preprocess(self):
...
 
             if SOOP_NRT_LOG_PATTERN.match(log_filename) is None:
                 raise InvalidFileNameError(
                     "SOOP NRT input logfile has incorrect naming '{name}'.".format(name=log_filename))
 
             # case to create netcdf files from log files
-            f.publish_type = PipelineFilePublishType.NO_ACTION
-
-            netcdf_filepath = netcdf_writer(f.src_path, self.temp_dir, self.ship_callsign_ls[SHIP_CODE])
+            f_log.publish_type = PipelineFilePublishType.NO_ACTION
```
bpasquer: Since we are generating a NetCDF from log files, I think we should archive the original log file now with the same path as on S3

aodndata/soop/soop_tmv_nrt.py
```diff
@@ -134,28 +174,44 @@ def preprocess(self):
...
+            if len(f_txt):
+                f_txt = f_txt[0]
+                f_txt.publish_type = PipelineFilePublishType.NO_ACTION
```
bpasquer: Same comment as above for the  original log files, probably best to archive them

aodndata/soop/soop_tmv_nrt.py
```diff
@@ -134,28 +174,44 @@ def preprocess(self):
...
 
-        f = self.file_collection[0]
+        f_txt = self.file_collection.filter_by_attribute_value('extension', '.txt')
+        f_log = self.file_collection.filter_by_attribute_value('extension', '.log')
+        f_nc = self.file_collection.filter_by_attribute_id('file_type', FileType.NETCDF)
 
-        if f.extension == '.nc':
+        if len(f_nc):
+            f_nc = f_nc[0]
             # case where we re-push an existing netcdf file
```
bpasquer: not quite sure when we would re-push netcdf files?

lbesnard: That can happen. to do some re-processing. I've done it already

aodndata/soop/soop_tmv_nrt.py
lbesnard: the archive is done here actually 
https://github.com/aodn/python-aodndata/pull/117/files/111cac9f339e4a3e97bf7dbd2ed338833c1103e1#diff-ab86673366e9f8fc8689d0db161fade2R208

the data provider only sends zip files. He doesn't send log files at all

PR #120, opened 2019-03-22
SOOP CO2 - Fix unittests, removing unacessary call to ship_callsign. …

lbesnard:
…Mod global attrs


lbesnard:
@lwgordonimos this should fix the unittests

PR #123, opened 2019-03-22
Code refactor

lbesnard:


lbesnard:
@lwgordonimos I did this for you ! improved by  +1.29%

aodndata/aims/faimms.py
```diff
@@ -52,20 +52,18 @@ def get_faimms_site_name(filepath):
...
-            return 'Heron_Island'
+        site_codes = {'DAV': 'Davies_Reef',
+                      'OI': 'Orpheus_Island',
+                      'OTI': 'One_Tree_Island',
+                      'RIB': 'Rib_Reef',
+                      'MRY': 'Myrmidon_Reef',
+                      'LIZ': 'Lizard_Island',
+                      'HI': 'Heron_Island'}
+
+        main_site_code = [site for site in site_codes.keys() if site in site_code]
```
lwgordonimos: This could just be:

```python
return site_codes.get(site_code, None)
```
You should always explicitly return something. Also, if you allow this to return None, then the other code has to handle it. Should this raise an Exception if the site code is invalid?

aodndata/soop/soop_trv.py
```diff
@@ -28,29 +32,16 @@ def preprocess(self):
...
 
-        if 'CPHL' in variables:
-            return 'CPHL'
-        elif 'TEMP' in variables:
-            return 'TEMP'
-        elif 'PSAL' in variables:
-            return 'PSAL'
-        elif 'TURB' in variables:
-            return 'TURB'
+        main_var = [var for var in nc_variables if var in self.var_dict.keys()]
```
lwgordonimos: Same here

aodndata/aims/faimms.py
lbesnard: for example 
```python
site_code = 'HIRP4'
```
this bit of code tests that "HI" from site_code is contained somewhere in the dictionarry site_codes

so  this would not work
 ```python
 site_codes.get(site_code, None)
```

the exception is raised in this other function
```python
def get_main_faimms_site_name_path(self, filepath):
```
but I guess you're right. It should be moved

aodndata/soop/soop_trv.py
lbesnard: also here nc_variables is an array of variables. It's trying to compare 2 arrays of string. so not sure how I could use the get

lbesnard:
@lwgordonimos could we merge this quick PR ?

PR #135, opened 2019-05-28
AIMS - correct regex to remove of md5 from filename in case there is …

lbesnard:
…no md5 in it

### DIFF
--- a/aodndata/aims/anmn_nrs.py
+++ b/aodndata/aims/anmn_nrs.py
@@ -1,10 +1,10 @@
 import os
 import re
-from datetime import datetime
-
-from netCDF4 import Dataset
 
 from aodncore.pipeline.handlerbase import HandlerBase
+
+from aodndata.aims.common import get_main_var_folder_name, get_year, get_product_version, get_site_code, \
+    remove_md5_from_filename
 
 
 class AnmnNrsAimsHandler(HandlerBase):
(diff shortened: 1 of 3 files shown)


aodndata/aims/anmn_nrs.py
```diff
@@ -56,7 +56,7 @@ def get_anmn_nrs_site_name(filepath):
 
     @staticmethod
     def remove_md5_from_filename(filepath):
-        return re.sub('.[0-9a-z]*.nc$', '.nc', filepath)
+        return re.sub('\.[0-9a-z]*.nc$', '.nc', filepath)
```
mhidas: Don't you want to escape the dot in ".nc" too? (i.e. "\.nc") Otherwise it means any character.
Also, since MD5 sums are always 32 characters in lenght, you could fix that in your regex too, i.e.
```regex
'\.[0-9a-z]{32}\.nc$'
```

mhidas:
Note that all Travis checks are failing due to boto3 package version issues (https://github.com/aodn/issues/issues/448)

lbesnard:
thanks @mhidas followed your suggestions

lbesnard:
@mhidas when you have some time

aodndata/aims/common.py
```diff
@@ -0,0 +1,63 @@
+import re
+from datetime import datetime
+
+from netCDF4 import Dataset
+
+
+def get_main_var(filepath):
+    netcdf_file_obj = Dataset(filepath, mode='r')
```
mhidas: This could also be a `with Dataset(filepath, mode='r') as netcdf_file_obj:`, as you've done in most  of your other functions.

aodndata/aims/common.py
```diff
@@ -0,0 +1,63 @@
...
+    variables = netcdf_file_obj.variables.keys()
+    netcdf_file_obj.close()
+
+    del variables[variables.index('TIME')]
+    del variables[variables.index('LATITUDE')]
+    del variables[variables.index('LONGITUDE')]
```
mhidas: This would be more straightforward:
```python
variables.remove('TIME')
```
etc..

mhidas:
Let me know if you want to make those tweaks I suggested. Otherwise I'm happy to merge this now.

lbesnard:
yeap doing them now. thanks

lbesnard:
@mhidas done

PR #136, opened 2019-05-30
Update Travis test matrix 

mhidas:
We want to test against the same pipeline stage (build/systest/prod) for all three artefacts aodndata is dependent on (aodncore, cc-plugin, ncwriter).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,20 +11,21 @@ env:
   global:
     - MPLBACKEND=agg
     - AODNFETCHER_URL="git+https://github.com/aodn/python-aodnfetcher.git@master"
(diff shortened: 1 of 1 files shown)


mhidas:
This will allow tests to partially succeed (at least in the `STAGE=build` case) when changes are made in aodndata that require a new version of one of the other artefacts that hasn't made it to prod yet. E.g. https://github.com/aodn/python-aodndata/pull/129

mhidas:
e.g. test result will look like this (https://travis-ci.org/aodn/python-aodndata/builds/539070447?utm_source=github_status&utm_medium=notification):
image

PR #143, opened 2019-06-20
SOOP TMV NRT - get correct measurement frequency of dataframe

lbesnard:


lbesnard:
Fix for issues with files having this kind of data in (i.e. missing latitude and longitude values):
https://github.com/aodn/python-aodndata/pull/143/files#diff-d0cde68f0e34df5420204fcc5e23a75cR3

PR #144, opened 2019-06-25
Exclude ADCP files from burst-averaging

mhidas:
The burst-processing code is only designed to operate on point timeseries data. The moorings handler was selecting files to process only based on the presence of  `instrument_burst_` global attributes, and an FV01 label.

It turns out ADCP files can have these attributes too (maybe erroneously - see https://github.com/aodn/imos-toolbox/issues/554), so we need to explicitly exclude them, because they  break the brust-averaging code.


mhidas:
@lwgordonimos I had forgotten about this PR. It should still be good to merge. Can we get it in before the Modernize stuff, or would you rather leave it until after?

mhidas:
@lbesnard @ocehugo This work-around is still needed (until https://github.com/aodn/imos-toolbox/issues/554 is fixed). We've just received another ADCP file with incorrect metadata and it prevented a whole deployment's worth of data being deployed.

If you have a spare minute today, please review & merge. Otherwise I'm happy to merge it myself.

test_aodndata/moorings/test_mooringsHandler.py
```diff
@@ -164,27 +166,28 @@ def test_all_formats_zip(self):
...
         f = handler.file_collection[0]
         self.assertEqual(f.name, BURST_NC_BASENAME)
         self.assertEqual(f.dest_path, os.path.join('IMOS/ANMN/NRS/NRSROT/Biogeochem_timeseries', BURST_NC_BASENAME))
-        self.assertTrue(f.is_checked)
-        self.assertTrue(f.is_stored)
-        self.assertTrue(f.is_harvested)
 
         # there should also be a burst-averaged file
-        f = handler.file_collection[1]
+        f = handler.file_collection[-1]
```
lbesnard: wouldn't it make more sense to use filtering option to get the file you need without using a specific index?

mhidas: I could, but this is simpler. Also, I know that the one I want is the last file added to the collection, because the burst-averaging happens after the resolve step. So if the last file in the collection _isn't_ the burst-averaged file, then there is something wrong, and the unittest _should_ fail.

mhidas:
I've just rebased this onto master to fix Travis tests (the old branch had Python3 compatibility issues)

PR #157, opened 2019-09-04
Moorings products handler

mhidas:
Prototype handler to create moorings products based on a JSON manifest file.

Depends on https://github.com/aodn/python-aodncore/pull/167 and https://github.com/aodn/python-aodncore/pull/169


aodndata/moorings/products_handler.py
```diff
@@ -0,0 +1,98 @@
...
+        if self.allowed_extensions is None:
+            self.allowed_extensions = ['.json']
+        self.product_site_code = None
+        self.product_variables = None
+
+    def preprocess(self):
+        """Collect available input files and create the products, adding them to the collection to be published."""
+
+        # Don't want to publish the manifest file, so remove it from the collection
+        self.file_collection.pop()
```
lwgordonimos: This could be avoided by making it a known file type(e.g. `.product_manifest`) and simply not adding to the collection in the first place

aodndata/moorings/products_handler.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Read the manifest file and extract key parameters for product
+        with open(self.input_file) as f:
```
lwgordonimos: Same here re: known file type, this validation could be moved to the resolve runner.

aodndata/moorings/products_handler.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+        # Find out what relevant input files are available on S3 for this site.
+        filter_list = [PropertyIsEqualTo(propertyname='site_code', literal=self.product_site_code),
+                       PropertyIsEqualTo(propertyname='file_version', literal='1'),
+                       PropertyIsEqualTo(propertyname='realtime', literal='false'),
+                       PropertyIsNotEqualTo(propertyname='data_category', literal='Biogeochem_profiles'),
+                       PropertyIsNotEqualTo(propertyname='data_category', literal='CTD_profiles')
+                       ]
+        filter = ogc_filter_to_string(And(filter_list))
```
lwgordonimos: `filter` is a reserved name, because of the builtin function

aodndata/moorings/products_handler.py
```diff
@@ -0,0 +1,98 @@
...
+        # TODO: replace ._wfs_broker.query_urls_for_layer() with .query_wfs_urls_for_layer() once aodncore has been updated
+        input_files = self.state_query._wfs_broker.query_urls_for_layer(self.FILE_INDEX_LAYER,
+                                                                        ogc_filter=filter,
+                                                                        url_property_name='url'
+                                                                        )
+        self.logger.info("Downloading {n} input files".format(n=len(input_files)))
+
+        # Download input files to local cache.
+        input_file_collection = RemotePipelineFileCollection(input_files)
+        input_file_collection.download(self._upload_store_runner.broker, self.temp_dir)
```
lwgordonimos: We need to consider a standard way to do this, so that it doesn't need to access the private `_upload_store_runner` property. Perhaps by adding method to HandlerBase, which abstracts this detail from child handlers.

e.g.

`self.download_collection(input_file_collection)`

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -0,0 +1,37 @@
...
+
+TEST_ROOT = os.path.dirname(__file__)
+GOOD_MANIFEST = os.path.join(TEST_ROOT, 'product_manifest.json')
+
+GETCAPABILITIES_FILE = os.path.join(TEST_ROOT, 'getCapabilities.xml')
+
+with open(GETCAPABILITIES_FILE) as f:
+    TEST_GETCAPABILITIES_RESPONSE = httpretty.Response(f.read())
+
+TEST_GETFEATURE_RESPONSE = httpretty.Response(
```
lwgordonimos: I think we should just put all these responses into separate files (despite me having inlined them in the first place!)

aodndata/moorings/products_handler.py
mhidas: :+1: Sure. I'd make it even more generic and call it `.json_manifest`, as it could be used for other purposes.
Or `.manifest.json`, so it has the "correct" extension for a JSON file?

aodndata/moorings/products_handler.py
mhidas: :+1: renamed to ogc_filter

aodndata/moorings/products_handler.py
mhidas: Yep, sounds good.

aodndata/moorings/products_handler.py
mhidas: Sure. The schema to validate against would still need to be defined here though, as it is specific to this handler.

mhidas: Ah, sorry. I was thinking of validating the content of the json, as below.

aodndata/moorings/products_handler.py
lwgordonimos: The only consideration here really is whether different JSON files should have different behaviour. For example, if in this case we *don't* want it in the collection, but in another case if a JSON file would have real data in it and actually be the "input file", we *would* want it in the collection.

aodndata/moorings/products_handler.py
lwgordonimos: Yeah was only thinking very high level stuff like "is it actually a valid JSON file". As you said, any further validation should be left up to the specific handler.

aodndata/moorings/products_handler.py
mhidas: As long as it's a _manifest_ file, we generally don't want it in the collection (the handler can add it later anyway).
But as you say, we would not want to do that for just _any_ JSON file, so we do need to distinguish the JSON manifest type from other JSON files.

test_aodndata/moorings/test_mooringsProductsHandler.py
mhidas: Done

aodndata/moorings/products_handler.py
lwgordonimos: Draft PR here: https://github.com/aodn/python-aodncore/pull/169

mhidas: Thanks, that seems to work.

mhidas:
@lwgordonimos This could do with another quick review if you have the time. I'll want to merge & test it in RC next week.

mhidas:
This is ready to merge & test on RC. Travis checks pass with the right versions of aodncore & aodntools: https://travis-ci.org/aodn/python-aodndata/builds/601128849

image

setup.py
```diff
@@ -55,7 +56,7 @@
     'cc-plugin-imos>=1.3.0',
     'matplotlib==1.5.1',
     'aodntools>=0.2.2',
-    'pandas==0.22.0'
+    'pandas==0.24.2'
```
lwgordonimos: Since it depends on a new aodncore feature, should also bump the required aodncore version as well to `>=0.31.0`

mhidas: Ah yes, good pick-up! Thanks. Will update.

mhidas: Done

PR #160, opened 2019-09-17
SRS SURFACE WAVE - separate dest_path for altimetry and scatterometry…

lbesnard:
… datasets


lbesnard:
just getting pipeline ready to handle new scatterometry dataset. will need to update chef-private databag to point altimetry to renamed dest_path function

PR #161, opened 2019-10-10
NSW-EOH - new pipeline

lbesnard:
@mhidas I'll need to chat with you regarding this pipeline to make sure I am doing what is expected


aodndata/nsw_oeh/handler.py
```diff
@@ -0,0 +1,434 @@
...
+        super(NswOehHandler, self).__init__(*args, **kwargs)
+
+        self.allowed_extensions = ['.zip']
+        self.campaign_folder_path = None
+
+    def preprocess(self):
+        self.file_collection.add(self.input_file_object)
+        file_collection = self.file_collection.filter_by_attribute_id('file_type', FileType.ZIP)
+        file_collection = file_collection[0]
+        file_collection.publish_type = PipelineFilePublishType.HARVEST_UPLOAD  # default
```
mhidas: This is unnecessarily convoluted. All you need to say is 
```python
self.input_file_object.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
```

aodndata/nsw_oeh/handler.py
```diff
@@ -0,0 +1,434 @@
...
+
+        pz = NSWOEHSurveyProcesor(file_collection._src_path)
```
mhidas:
```suggestion
        pz = NSWOEHSurveyProcesor(self.input_file)
```

aodndata/nsw_oeh/handler.py
```diff
@@ -0,0 +1,434 @@
...
+
+        return publish_files
+
+
+class NswOehHandler(HandlerBase):
+    def __init__(self, *args, **kwargs):
+        super(NswOehHandler, self).__init__(*args, **kwargs)
+
+        self.allowed_extensions = ['.zip']
+        self.campaign_folder_path = None
```
mhidas: Minor point, but `survey_path` would be more consistent with the terminology used in the rest of the code (and used by NSW OEH folks)

mhidas:
Also needs to be rebased onto master to resolve conflicts in setup.py

aodndata/nsw_oeh/handler.py
lbesnard: done

aodndata/nsw_oeh/handler.py
lbesnard: done

aodndata/nsw_oeh/handler.py
lbesnard: done

lbesnard:
>     * Publish the files as per the logic explained in the `NSWOEHSurveyProcesor.extract()` doc string ([line 386](https://github.com/aodn/python-aodndata/pull/161/files#diff-573d56f4b2c6d6ec3625e8698a6aea66R386) -- the method itself is no longer needed, as the handler has already unpacked the zip file)

I looked at your new unittests and debug them to understand what had to be done. I didn't understand the text under the extract() function, and what was supposed to happen clearly.

So please have a look at my changes, making sure I set up the right HARVEST_UPLOAD ... to the right files ...

passing all the tests anyway

mhidas:
> passing all the tests anyway

I'm surprised by this statement. I get:
```
  File "/sw/chef/src/python-aodndata/test_aodndata/nsw_oeh/test_nsw_oeh_handler.py", line 258, in test_good_mb
    self.assertEqual(10, len(file_collection_harvest_upload._PipelineFileCollection__s.item_list))
AttributeError: 'PipelineFileCollection' object has no attribute '_PipelineFileCollection__s'
```

Anyway, looking at it now...

test_aodndata/nsw_oeh/test_nsw_oeh_handler.py
```diff
@@ -0,0 +1,299 @@
...
+
+        with zipfile.ZipFile(GOOD_STAX_ZIP) as zf:
+            expected_files = {os.path.basename(p) for p in zf.namelist() if '_STAX_SHP.' in p}
+        expected_files.add(os.path.basename(GOOD_STAX_ZIP))
+        expected_dest_paths = {os.path.join(dest_path_base, p) for p in expected_files}
+        result_dest_paths = {f for f in handler.file_collection.get_attribute_list('dest_path') if f}
+
+        self.assertSetEqual(expected_dest_paths, result_dest_paths)
+        self.assertEqual(1, len(handler.file_collection.filter_by_attribute_id('publish_type',
+                                                                               PipelineFilePublishType.UPLOAD_ONLY).
```
mhidas: Both the shapefile and the original zip should be HARVEST_UPLOAD.

aodndata/nsw_oeh/handler.py
```diff
@@ -0,0 +1,432 @@
...
+
+        if len(report) > 0:
+            raise_string = ''
+            for heading, messages in report.items():
+                raise_string += "\n{heading}:\n{messages}\n".format(
+                    heading=heading,
+                    messages="\n".join(["{0}. {1}".format(i+1, messages) for i, messages in enumerate(messages)]))
+
+            raise InvalidFileContentError(
+                "Could not determine data category for '{name}'\n{raise_string}".format(
```
mhidas: The error message should be something like
```suggestion
                "Zip file '{name}' failed content checks:'\n{raise_string}".format(
```

aodndata/nsw_oeh/handler.py
```diff
@@ -0,0 +1,432 @@
...
+                    name=os.path.basename(self.input_file),
+                    raise_string=raise_string))
+
+        # stax files
+        if pz.survey_methods in BASIC_PACKAGED_METHODS:
+            self.file_collection.set_publish_types(PipelineFilePublishType.NO_ACTION)  # reset all publish types
+
+            self.file_collection.filter_by_attribute_id('file_type', FileType.ZIP).\
+                set_publish_types(PipelineFilePublishType.UPLOAD_ONLY)  # publish zip file
```
mhidas: Shouldn't base this on the just file_type, because there may be additional zip files _inside_ the overall uploaded zip file. Use `self.input_file_object` instead.
Also, it should be `HARVEST_UPLOAD` here too.

lbesnard:
> > passing all the tests anyway
> 
> I'm surprised by this statement. I get:

well, believe it or not, I just tried debugging this and this unittest is passing on my machine (proper working env too). It's really bizarre

lbesnard:
@lwgordonimos it seems like travis is not happy because of the test using
assertCountEqual which is only for python3. see https://docs.python.org/2/library/unittest.html

It would be good to only use python3 in travis

mhidas:
> well, believe it or not, I just tried debugging this and this unittest is passing on my machine (proper working env too). It's really bizarre

I think the reason the unittests were not failing before is that they were not being run at all. My guess is that this was because test_aodndata/nsw_oeh/ was missing an `__init__.py` (added in b7212128d27828a1b73276315231f031a2be8b52).

mhidas:
Good pick-up on the fact that aodndata/nsw_oeh/ was also missing an init file. I would have though the unittests should have failed because of that too...

lbesnard:
> I think the reason the unittests were not failing before is that they were not being run at all. My guess is that this was because test_aodndata/nsw_oeh/ was missing an `__init__.py` (added in [b721212](https://github.com/aodn/python-aodndata/commit/b7212128d27828a1b73276315231f031a2be8b52)).

well no actually, I was running this test individually on pycharm so I'm a bit puzzled

mhidas:
> It would be good to only use python3 in travis

Yes! See #180 

Until that is merged, we just need to look at the [build job details](https://travis-ci.org/aodn/python-aodndata/builds/613874934?utm_source=github_status&utm_medium=notification) on Travis. Even though the overall status was a fail, we can see it passes for Python 3.5 & 3.6.

lbesnard:
shall we merge?

mhidas:
Thanks @lbesnard

mhidas:
Bugger... this also fails to [build](https://jenkins.aodn.org.au/view/aodndata/job/aodndata_build/133/) on Jenkins because of the same Python 2 incompatibility in the unittests!

@lwgordonimos Is this on the python migration roadmap?

PR #162, opened 2019-10-17
SOOP vocab - handle case of new vessel having similar name as previou…

lbesnard:
…s vessel, but new callsign


lbesnard:
same as https://github.com/aodn/data-services/pull/956

PR #167, opened 2019-10-22
Update handlers using state_query to use keys method, for compatibili…

lwgordonimos:
…ty with upcoming change to use RemotePipelineFileCollection as return type

This is to ensure the handlers will cope with: https://github.com/aodn/python-aodncore/pull/167, as the current assumption is that the return type of `self.state_query.query_storage()` is a "dict of dicts".


mhidas:
@bpasquer @lbesnard FYI - I'd like to merge this asap as it is a prerequisite for the aodncore changes I need for the moorings products handler.

PR #178, opened 2019-11-19
Moorings products: add download/opendap URL details

mhidas:
Add more useful metadata to the aggregated timeseries products generated, using the additional arguments introduced in https://github.com/aodn/python-aodntools/pull/78

(Also a bit of code clean-up.)


mhidas:
I've scaled this PR back to just the key change I want to make (download URL details). I'll do the refactoring later.

mhidas:
Travis checks only fail because the required version of aodntools hasn't been promoted to prod yet. The "build" and "systest" versions do pass (see https://travis-ci.org/aodn/python-aodndata/builds/613788501?utm_medium=notification&utm_source=github_status)

PR #179, opened 2019-11-19
Stop doing Travis tests in python 2.7

mhidas:
We are now running this code under python 3.5 everywhere, so we shouldn't need to test under 2.7 anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.7"
   - "3.5"
   - "3.6"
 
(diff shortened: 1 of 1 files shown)


lwgordonimos:
Was thinking it might be best to include this in an overall "drop PY2" PR: https://github.com/aodn/python-aodndata/pull/180

I was playing with one for aodncore, and had a look at data as well (and there is much less to do here): https://github.com/aodn/python-aodncore/pull/183

mhidas:
:+1: Yep, I had a hunch you'd say that. Makes sense.

Can we add a7216373329ed4f52d5129b6b5a15171be6b80ce to that PR? I realised the "prod artefacts" config is exactly the same as the basic "Python 3.5" one.

PR #180, opened 2019-11-19
Remove Python 2 compatibility code

lwgordonimos:
~~IMPORTANT: this is not to be merged until the build job is ported to new Jenkins.~~


setup.py
```diff
@@ -56,8 +56,7 @@
     'aodncore>=0.31.0',
     'cc-plugin-imos>=1.3.0',
     'matplotlib==1.5.1',
-    'aodntools>=0.4.3',
-    'pandas==0.24.2'
```
mhidas: We may not need to pin the version, but we _are_ using pandas in several handlers, so we should keep it here.

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.7"
   - "3.5"
```
mhidas: This gives us a "Python 3.5" environment, which by default uses prod artefacts, so we can actually remove this entry under `matrix` below:
```yaml
     - name: "prod artefacts"
      env: STAGE="prod"
```

mhidas:
Maybe when we get this merged, it would be a good excuse to finally update the version to 1.0!

lwgordonimos:
Yes, sounds like a plan!

setup.py
lwgordonimos: I'm trying to find documentation of the best practice approach for this (importing dependencies of dependencies), but removing it is in line with several other key packages, namely `compliance-checker` (installed by aodncore), `numpy` (installed by `compliance-checker`, so 2 "hops" away in the dependency chain) and `netCDF4` (installed by `compliance-checker`).

So basically all of the scientific libraries which are installed indirectly by aodncore, aodntools and cc-plugin-imos. Dependency clashes are a huge pain, so I was hoping for the simplest approach here. If we *know* that aodntools and aodncore install the scientific libs, I can't think of a good reason to explicitly duplicate the dependency here?

mhidas: My thinking was that we should have `pandas` listed here because it is explicitly imported by some of the handlers, unlike `compliance-checker`, which is only used indirectly via `aodncore`.
But `numpy` and `netCDF4` are also imported directly and not listed here as a dependency, which is ok.

So I agree, let's try to avoid dependency clashes.

mhidas:
The Jenkins build also needs to be updated to Python 3 before we can merge this - see https://github.com/aodn/issues/issues/595

mhidas:
Also need to revert 3fe7cb07cf711dc5b5dace842f73ad799e3819ee

mhidas:
41147d6bef1fdf53d7ff07c55f9a06414ec147f0 will need reverting too if it gets merged before this PR.

mhidas:
I think this needs a rebase. The versions are inconsistent between [bumpversion.cfg](https://github.com/aodn/python-aodndata/blob/remove-py2/.bumpversion.cfg#L2) and [version.py](https://github.com/aodn/python-aodndata/blob/remove-py2/aodndata/version.py)

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,25 @@
+<component name="InspectionProjectProfileManager">
```
mhidas: Something weird going on here... none of these files have anything to do with .travis.yml ...
And I'm sure you didn't mean to commit any wheels in here!

lwgordonimos:
@mhidas this should be right now, but travis fails due to the aodncore production stage not meeting the `>=1.0.0` requirement until that is promoted. We should promote the other first, verify this one and then merge this one.

setup.py
```diff
@@ -55,15 +53,15 @@
 }
 
 INSTALL_REQUIRES = [
-    'aodncore>=0.31.0',
+    'aodncore>=1.0.0',
     'cc-plugin-imos>=1.3.0',
     'matplotlib==1.5.1',
     'aodntools>=0.4.4',
```
mhidas: Ooops, one last thing: should we change this dependency to `aodntools>=1.0.0` too?

PR #186, opened 2019-12-05
Moorings products: delete previous versions and refactoring

mhidas:


mhidas:
The last commit (31facf5b12d5190d8991530c6ba22a4b9863dd39) requires https://github.com/aodn/python-aodncore/pull/188, included in aodncore 0.31.5

mhidas:
I've rebased onto master and dropped the Python2 compatibility commit, since we're no longer running or testing in Python 2 anywhere.

mhidas:
Ready for review.

Once we get this in prod we can start regularly updating the aggregated timeseries product (and then we can add the others too).

aodndata/moorings/products_handler.py
```diff
@@ -43,25 +43,31 @@ def __init__(self, *args, **kwargs):
         self.allowed_extensions = ['.json_manifest']
         self.product_site_code = None
         self.product_variables = None
+        self.input_file_collection = None
+        self.input_file_variables = None
+        self.excluded_files = defaultdict(set)
```
ocehugo: I don't see a reason to use defaultdict here since the dict is very simple - explicit is better than implicit.

aodndata/moorings/products_handler.py
```diff
@@ -43,25 +43,31 @@ def __init__(self, *args, **kwargs):
...
 
-    def preprocess(self):
-        """Collect available input files and create the products, adding them to the collection to be published."""
-
-        # Read the manifest file and extract key parameters for product
+    def _read_manifest(self):
+        """Read the manifest file and extract key parameters for product"""
         with open(self.input_file) as f:
             manifest = json.load(f)
+
```
ocehugo: I would avoid try here and ask explicitly the keys - faster and cleaner.
```
mkeys = manifest.keys();
invalid_file = 'site_code' not in mkeys and 'variables' not in mkeys;
if invalid_file:
```

aodndata/moorings/products_handler.py
```diff
@@ -79,23 +84,21 @@ def preprocess(self):
                                                                     filter=ogc_filter,
                                                                     propertyname=['url', 'variables']
                                                                     )
```
ocehugo: I would explicitly check for wfs_response['features'] being not empty instead of using hidden logic in a dict comprehension using a for loop. Anyway, you need to bail out eventually if no features are available - which the code is not doing.

```
url = f['properties']['url']
vars = f['properties']['variables'].split(', ')
feat = wfs_response['features']
if feat:
	self.input_file_variables = {url:vars};
else:
	raise...
```

aodndata/moorings/products_handler.py
```diff
@@ -79,23 +84,21 @@ def preprocess(self):
...
+        self.input_file_collection.download(self._upload_store_runner.broker, self.temp_dir)
         # TODO: Replace temp_dir above with cache_dir?
 
-        # TODO: Run compliance checks and remove non-compliant files from the input list (log them).
+    def _make_aggregated_timeseries(self):
+        """For each variable, generate product and add to file_collection."""
 
-        # For each variable, generate product and add to file_collection.
-        self.excluded_files = defaultdict(set)
         for var in self.product_variables:
```
ocehugo: faster,shorted and more explicit:

```
p_vars = self.product_variables
fitems = self.input_file_variables.items()
input_list = [file for file,vlist in fitems for var in p_vars if var in vlist]
```

aodndata/moorings/products_handler.py
```diff
@@ -116,6 +119,20 @@ def preprocess(self):
             product_file.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
             self.file_collection.add(product_file)
 
+    def preprocess(self):
```
ocehugo: just rename to ```process``` - it's making stuff in make_aggregate_timeseries.

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -50,6 +50,10 @@ def test_good_manifest(self):
         upload_broker.upload(INPUT_FILE_COLLECTION)
 
         handler = self.run_handler(GOOD_MANIFEST)
+        self.assertCountEqual(INPUT_FILE_COLLECTION.get_attribute_list('dest_path'),
```
ocehugo: the inputs are so long that they occupy the same amount of lines of assigning pointers to the individual variables.

```
e_path = INPUT_FILE_COLLECTION.get_attribute_list('dest_path')
r_path = handler.input_file_collection.get_attribute_list('dest_path')
self.assertEqual(e_path,r_path)
```

aodndata/moorings/products_handler.py
```diff
@@ -60,6 +60,26 @@ def _read_manifest(self):
                 "manifest file '{self.input_file}' missing information (site_code, variables)".format(self=self)
             )
 
+    def _get_wfs_features(self, filter_list, propertyname='*'):
```
ocehugo: I tend to avoid using "internal" names to functions when they don't change state of the class - the function apparently only extract a piece of information without changing any state. Hence, it should be "public" and without the leading ```_```.

aodndata/moorings/products_handler.py
```diff
@@ -75,19 +95,12 @@ def _get_input_files(self):
...
-        # Note I need to access _wfs_broker to be able to use query_urls_for_layer() with a filter,
-        # as the corresponding StateQuery method doesn't accept additional kwargs.
-        # TODO: find out why this calls getCapabilities twice (and takes 40s even when response mocked with httpretty)
-        # TODO: replace ._wfs_broker.getfeature_dict() with .getfeature_dict() once aodncore has been updated
-        wfs_response = self.state_query._wfs_broker.getfeature_dict(typename=[self.FILE_INDEX_LAYER],
-                                                                    filter=ogc_filter,
-                                                                    propertyname=['url', 'variables']
-                                                                    )
+        wfs_features = self._get_wfs_features(filter_list, propertyname=['url', 'variables'])
         self.input_file_variables = {f['properties']['url']: f['properties']['variables'].split(', ')
```
ocehugo: Same comment as the one before - just check for emptyness of wfs_features explicitly.

aodndata/moorings/products_handler.py
lwgordonimos: That is changing from EAFP to LBYL, which is not idiomatic:

https://en.wikipedia.org/wiki/Python_syntax_and_semantics#Exceptions
https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

aodndata/moorings/products_handler.py
```diff
@@ -106,6 +108,31 @@ def _get_input_files(self):
...
+    def _get_old_product_files(self):
+        """Get a list of the currently published aggregated_timeseries files for the site being processed."""
+
+        filter_list = [PropertyIsEqualTo(propertyname='site_code', literal=self.product_site_code),
+                       PropertyIsEqualTo(propertyname='data_category', literal='aggregated_timeseries')
+                       ]
+        wfs_features = self._get_wfs_features(filter_list, propertyname=['url'])
+
+        self.old_product_files = {}
+        var_pattern = re.compile(r'FV01_([A-Z0-9-]+)-aggregated')
```
ocehugo: Apparently this can be promote to a global - "AGG_VAR_PATTERN" or something.

aodndata/moorings/products_handler.py
```diff
@@ -106,6 +108,31 @@ def _get_input_files(self):
...
+        for f in wfs_features:
```
ocehugo: maybe better to report all files not just the first one? Same number of lines but doing more :)

```
urls = [x['properties']['url'] for x in wfs_features];
emsg =  "Could not determine variable of interest for "
invalids = [emsg+f"{url} for url in urls if var_pattern.search(url)]
if invalids:
    raise InvalidFileNameError('\n'.join(invalids))
else: 
    vmatches = [var_pattern.search(url) for url in urls]
    variables_of_interest = [var.group(1).replace('-','_') for var in vmatches]
```

aodndata/moorings/products_handler.py
```diff
@@ -132,6 +159,14 @@ def _make_aggregated_timeseries(self):
             product_file.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
             self.file_collection.add(product_file)
```
ocehugo: I would not expect that a "make" method would delete files. I think it's better to be explicit in the call to remove - maybe a cleanup method or renaming here.

aodndata/moorings/products_handler.py
ocehugo: IMO, EAFP is overused for simple cases such as the above.

You use more lines, refactoring to list comprehensions (or generators) takes more time later, etc.

TO call that more "idiomatic" is a thin line - Why I should fall back to exception handling if I know what I'm looking for? If it was a function call wrap I would agree, but it's a simple comparison in standard classes.

aodndata/moorings/products_handler.py
mhidas: Ok, I've factored out the "delete" code in a new method, but it's still easiest to call it here, where the new product just got created, as I have all the information I need.
Note that the actual publishing and deleting happens in a later `publish` step, based on the attributes of each file in the `file_collection` (see https://aodn.github.io/python-aodncore/userdoc.html)

aodndata/moorings/products_handler.py
mhidas: Again, this is overriding a method in `HandlerBase` so that it is executed at a specific step in the handler workflow. See https://github.com/aodn/python-aodncore/blob/master/HandlerBase_state_machine.png and https://aodn.github.io/python-aodncore/userdoc.html (note the documentation is a bit out of date, but the overview is still correct - for more details see the [handlerbase](https://github.com/aodn/python-aodncore/blob/master/aodncore/pipeline/handlerbase.py) code)

mhidas:
Rebased onto master again, hopefully this will fix all the Travis tests...

aodndata/moorings/products_handler.py
ocehugo: I mean why you should set a empty dictionary with a dictionary comprehensions with a for loop? Just check if was features is empty and raise the exception...

aodndata/moorings/products_handler.py
ocehugo: Yeap, but despite being only used within the private func,  it's a data condition not method. In other words, the nature of the definition can lead to re-use in the future.

aodndata/moorings/products_handler.py
mhidas: Ok. Check added.

aodndata/moorings/products_handler.py
mhidas: Ok. done

aodndata/moorings/products_handler.py
mhidas: Done

mhidas:
@ocehugo I've made a couple more changes you requested. I can't spend any more time on this right now, but would like to run this in prod by the end of the week. 
Please merge so I can test it in RC.

ocehugo:
just merged - I don't know why github lost the track of some comments since the forced push.

mhidas:
Thanks.
I thin that happens when the specific line/change you comment on is changed by a force push.

PR #191, opened 2019-12-11
SOOP XBT - fix corrupted jpg

lbesnard:


aodndata/soop/soop_xbt_dm.py
```diff
@@ -117,8 +117,9 @@ def create_plot(netcdfFilePath, output_dir):
                   lon="%0.2f" % lon,
                   date_start=date_start))
 
-    jpg_output = tempfile.NamedTemporaryFile(delete=False, dir=output_dir)
-    savefig(jpg_output)
+    img_output = tempfile.NamedTemporaryFile(delete=False, dir=output_dir)
+    savefig(img_output)
+    Image.open(img_output.name).convert('RGB').save(img_output.name, 'JPEG', quality=80, optimize=True, progressive=True)  # convert png to jpg
```
mhidas: This seems a bit of a roundabout way of doing it. Not sure what version of matplotlib is required, but the [docs for 3.1.1](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.savefig.html) seem to imply you can just use `savefig` to write a JPG.

This [discussion thread](https://stackoverflow.com/questions/8827016/matplotlib-savefig-in-jpeg-format) (a few years old - see second answer) you still need to install [`pillow`](https://pypi.org/project/Pillow/), which is the updated fork of PIL (which itself is no longer supported, i guess?)

setup.py
```diff
@@ -56,8 +56,9 @@
     'aodncore>=1.0.0',
     'aodntools>=1.0.0',
     'cc-plugin-imos>=1.3.0',
-    'matplotlib>=1.5.1',
-    'fiona>=1.8.8'
+    'fiona>=1.8.8',
+    'image',
```
mhidas: According to PyPI, [image](https://pypi.org/project/image/) is something else. I think you want `pillow` (see my other comment).

aodndata/soop/soop_xbt_dm.py
lbesnard: yeah this is what I followed

lbesnard: putting a new PR. yeah you're right, the new matplotlib version can create a jpg

setup.py
lbesnard: so this all PR works either with image or pillow.
If I do it with pillow, the code gets kind of simpler, however pillow is never directly imported. So one might wonder in the future why we put pillow in the setup.py
If I do it with image, it is imported directly.

mhidas: I still think `image` is not what you want ([PyPI](https://pypi.org/project/image/) says it's a Django application). It just happens to depend on `pillow`...
```
(python-aodndata-virtualenv) mhidas@adele2:/sw/chef/src/python-aodndata$ pip install image
Collecting image
  Downloading https://files.pythonhosted.org/packages/0c/ec/51969468a8b87f631cc0e60a6bf1e5f6eec8ef3fd2ee45dc760d5a93b82a/image-1.5.27-py2.py3-none-any.whl
Collecting django
  Downloading https://files.pythonhosted.org/packages/43/d6/0aed0b12c66527748ce5a007da4618a65dfbe1f8fca82eccedf57d60295f/Django-3.0-py3-none-any.whl (7.4MB)
     |████████████████████████████████| 7.4MB 6.4MB/s 
Collecting pillow
  Downloading https://files.pythonhosted.org/packages/10/5c/0e94e689de2476c4c5e644a3bd223a1c1b9e2bdb7c510191750be74fa786/Pillow-6.2.1-cp36-cp36m-manylinux1_x86_64.whl (2.1MB)
     |████████████████████████████████| 2.1MB 3.1MB/s 
Requirement already satisfied: pytz in ./python-aodndata-virtualenv/lib/python3.6/site-packages (from django->image) (2019.3)
Collecting asgiref~=3.2
  Downloading https://files.pythonhosted.org/packages/a5/cb/5a235b605a9753ebcb2730c75e610fb51c8cab3f01230080a8229fa36adb/asgiref-3.2.3-py2.py3-none-any.whl
Collecting sqlparse>=0.2.2
  Downloading https://files.pythonhosted.org/packages/ef/53/900f7d2a54557c6a37886585a91336520e5539e3ae2423ff1102daf4f3a7/sqlparse-0.3.0-py2.py3-none-any.whl
Installing collected packages: asgiref, sqlparse, django, pillow, image
Successfully installed asgiref-3.2.3 django-3.0 image-1.5.27 pillow-6.2.1 sqlparse-0.3.0
```

I don't think we need all that excess baggage just to convert an image to jpg.

mhidas: If you want to make it clear why you're adding the dependency, add a comment to explain it (in setup.py, and possibly also in the SOOP XBT code where you rely on this package)

lbesnard:
@mhidas FYI, there's a new commit

lwgordonimos:
FYI I've added this PR to avoid the `magic` requirement here, since it's something I've been meaning to add to aodncore anyway: https://github.com/aodn/python-aodncore/pull/194

setup.py
```diff
@@ -56,8 +56,10 @@
     'aodncore>=1.0.0',
     'aodntools>=1.0.0',
     'cc-plugin-imos>=1.3.0',
+    'fiona>=1.8.8',
     'matplotlib>=1.5.1',
-    'fiona>=1.8.8'
+    'pillow',  # indirectly used by matplotlib in savefig. See soop_xbt_dm handler
+    'python-magic'
```
mhidas: If you're only using this in the unittests, it should go under `TESTS_REQUIRE`, below.

test_aodndata/soop/test_soop_xbt_dm.py
```diff
@@ -18,20 +19,26 @@ def setUp(self):
...
                              GOOD_NC)
                          )
-        self.assertTrue(f.is_checked)
-        self.assertTrue(f.is_stored)
+        self.assertTrue(f_nc.is_checked)
+        self.assertTrue(f_nc.is_stored)
+
+        f_jpg = handler.file_collection[1]  # cant filter by extension because this is a tempfile generated file until dest_path is set
+        image_path = os.path.join(handler.state_query._storage_broker.prefix,
+                                  f_jpg.dest_path)
```
mhidas: Can't you just access the file locally at `f_jpg.local_path`?

setup.py
lbesnard: i removed this because it's going to be in the new core, but yeah you're right

test_aodndata/soop/test_soop_xbt_dm.py
lbesnard: no, at this stage it is already uploaded as the handler is creating it.

lbesnard:
good to go now

PR #193, opened 2019-12-12
SOOP BA - fix corrupted png breaking unittests

lbesnard:


mhidas:
:+1: Thanks @lbesnard !

Issue #194, opened 2019-12-13
soop_co2_rt_nc_generator unable to process some nan values 


bpasquer:
The process to extract data and generate NetCDF from  SOOP COR RT text files sometimes output the following error: `ValueError: Unable to parse string "   nan"`
The conversion of `nan` to numeric fails.

This is the full log
>Traceback (most recent call last):
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/aodncore/pipeline/handlerbase.py", line 1037, in run
    self.trigger(transition['trigger'])
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 65, in _get_trigger
    return machine.events[trigger_name].trigger(model, *args, **kwargs)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 405, in trigger
    return self.machine._process(func)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 1073, in _process
    return trigger()
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 423, in _trigger
    return self._process(event_data)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 433, in _process
    if trans.execute(event_data):
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 279, in execute
    machine.callback(func, event_data)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/transitions/core.py", line 1031, in callback
    func(*event_data.args, **event_data.kwargs)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/aodndata/soop/soop_co2.py", line 97, in preprocess
    nrt_nc_file_path = soop_co2_nrt_nc_generator.process_co2_rt(rt_file, self.products_dir, self.ship_callsign_ls)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/aodndata/soop/soop_co2_nrt_nc_generator.py", line 64, in process_co2_rt
    (dataf, platform_code) = read_realtime_file(realtime_file)
  File "/mnt/ebs/pipeline/lib/python3.5/site-packages/aodndata/soop/soop_co2_nrt_nc_generator.py", line 224, in read_realtime_file
    input_rt_parameter, self.src_path)
[...]

PR #158, opened 2019-09-11
SOOP-CO2 NRT: Remove call to coerce function to make code Python 3 co…

bpasquer:
…mpatible

PR #196, opened 2019-12-13
SOOP CO2 - fix issue with txt file containing NaNs and whitespace

lbesnard:


lbesnard:
@bpasquer  fixing https://github.com/aodn/python-aodndata/issues/194

bpasquer:
@lbesnard my worry is that it is conflicting with https://github.com/aodn/python-aodndata/pull/158

PR #197, opened 2019-12-16
Update moorings product handler + test in line with changes to aodnco…

lwgordonimos:
…re in https://github.com/aodn/python-aodncore/pull/199


mhidas:
:+1: This looks great, and the tests run much faster.

The only problem is that both the handler and the unittests have change on master, and it's a bit of an awkward merge. I'll see if I can figure it out...

mhidas:
@lwgordonimos This all looks good to me, and the tests are all passing. Can we merge it?

mhidas:
It would then allow us to switch aodncore to use WFS 1.1.0 by default (as per https://github.com/aodn/python-aodncore/pull/199#issuecomment-566305135)

lwgordonimos:
@mhidas it needed a rebase again due to the changes made since then, and I noticed the fixup commit only had IDE metadata and WHL files in it so dropped it. Aside from that, I think it's pretty good to go now.

mhidas:
:+1: Sweet!

Travis test is failing for a totally unrelated reason: The Python 3.6 environment now installs the just-released [Pandas 1.0.0](https://pandas.pydata.org/docs/whatsnew/v1.0.0.html), causing three SOOP CO2 tests to fail. I guess we'll worry about that when we upgrade to Python 3.6 or Pandas 1?

lwgordonimos:
It may be a side effect of this: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.0.html#dedicated-string-data-type

I found this suggestion so added a commit to try it out: https://stackoverflow.com/a/52065957

PR #198, opened 2019-12-17
Update wfs-changes branch with changes from master

mhidas:
I had a go, but needs some more work to make the unittests pass...


test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -13,15 +12,14 @@
...
-with open(GETCAPABILITIES_FILE) as f:
-    TEST_GETCAPABILITIES_RESPONSE = httpretty.Response(f.read())
+GETFEATURE_OLD_PRODUCTS_FILE = os.path.join(TEST_ROOT, 'getFeature_old_products.json')
 
 with open(GETFEATURE_FILE) as f:
     TEST_GETFEATURE_JSON = f.read()
-TEST_GETFEATURE_RESPONSE = httpretty.Response(TEST_GETFEATURE_JSON)
+
+with open(GETFEATURE_OLD_PRODUCTS_FILE) as f:
+    TEST_GETFEATURE_OLD_PRODUCTS_JSON = f.read()
```
mhidas: Because the handler is now doing TWO getfeature requests, we need to mock this second one as well. How do I do that?

lwgordonimos: Use `side_effect` rather than return value. if you set side_effect to an Iterable, it will return the elements in sequence when the mock is called repeatedly:

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

mhidas:
Thanks, that works great!

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.3
+current_version = 1.0.5
```
lwgordonimos: Looks like the rebase has pulled in some other commits? Shouldn't be any bumpversion things here.

aodndata/soop/soop_co2_nrt_nc_generator.py
```diff
@@ -222,6 +222,7 @@ def read_realtime_file(self):
 
     dataf = check_parameters(dataf, platform_code_short,
                              input_rt_parameter, self.src_path)
+    dataf = dataf.apply(lambda x: x.str.strip() if x.dtype == "object" else x)  # strip whitespace
```
lwgordonimos: Looks like a commit from another branch

PR #205, opened 2020-01-24
Cleanup matplotlib objects to avoid excessive memory growth

lwgordonimos:


lbesnard:
this is a huge decrease of code coverage @lwgordonimos . I'm not impressed

PR #206, opened 2020-02-14
Create hourly product in moorings products pipeline

mhidas:


aodndata/moorings/products_handler.py
```diff
@@ -219,16 +230,16 @@ def _make_aggregated_timeseries(self):
             product_file.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
             self.file_collection.add(product_file)
 
-            self._cleanup_previous_version(product_file.name, var)
+            self._cleanup_previous_version(product_file.name)
```
ocehugo: Nice refact above.

aodndata/moorings/products_handler.py
```diff
@@ -219,16 +230,16 @@ def _make_aggregated_timeseries(self):
...
+        """Delete any previously published version(s) of the given product file.
         Ignores cases where the previous version has exactly the same file name, as this will simply be overwritten.
 
         :param product_name: Name of the newly generated product
-        :param var: Name of the variable of interest
         """
-        for old_product_url in self.old_product_files.get(var, []):
+        product_type = get_product_type(product_name)
+        for old_product_url in self.old_product_files.get(product_type, []):
             if os.path.basename(old_product_url) != product_name:
```
ocehugo: IMO, the function intent is not clear by the directives within.
examples:

1.How ```old_product_url``` will ever match ```product_name```?  Maybe this difference if is better expressed with other 
names or the expression is better represented by a meaningful variable name ?

2.Also, I don't understand why we are ```self.file_collection.add(old_file)```  if we want to cleanup older ones. 

my point is, or the code above is not doing what it should be or the intent appears hidden in the implementation of other functions.

aodndata/moorings/products_handler.py
```diff
@@ -111,6 +112,7 @@ def __init__(self, *args, **kwargs):
         self.input_file_collection = None
         self.input_file_variables = None
         self.excluded_files = dict()
+        self.product_qc_flags = [[1, 2], [0, 1, 2]]
```
ocehugo: maybe a comment about why the nested lists or using a dictionary?

aodndata/moorings/products_handler.py
```diff
@@ -180,10 +182,13 @@ def _get_input_files(self):
         # TODO: Replace temp_dir above with cache_dir?
 
     def _get_old_product_files(self):
-        """Get a list of the currently published aggregated_timeseries files for the site being processed."""
+        """Get a list of the currently published product files for the site being processed."""
 
         filter_list = [PropertyIsEqualTo(propertyname='site_code', literal=self.product_site_code),
```
ocehugo: since this is very chatty, I would create two simple variables:
```
match_site_code = PropertyIsEqualto(...)
match_product_category = Or([...])
filter_list = [match_site_code, match_products_category]
```

aodndata/moorings/products_handler.py
```diff
@@ -232,6 +237,31 @@ def _make_aggregated_timeseries(self):
...
+    def _make_hourly_timeseries(self):
+        """Generate hourly products for the site and add to file_collection."""
+
+        # Filter input_list to the files relevant for this var
+        input_list = [f.local_path for f in self.input_file_collection]
+        self.logger.info("Creating hourly products from {n} input files".format(n=len(input_list)))
+
+        for qc_flags in self.product_qc_flags:
+
+            product_url, errors = hourly_aggregator(input_list, self.product_site_code, qc_flags, self.temp_dir)
```
ocehugo: I assume product_url is different given the ```qc_flags``` arguments!?

aodndata/moorings/products_handler.py
```diff
@@ -265,6 +295,7 @@ def preprocess(self):
         # TODO: Run compliance checks and remove non-compliant files from the input list (log them).
```
ocehugo: Why when you intent to cleanup you are actually creating more stuff?

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -54,19 +54,32 @@ def test_good_manifest(self, mock_webfeatureservice):
         self.assertCountEqual(INPUT_FILE_COLLECTION.get_attribute_list('dest_path'),
                               handler.input_file_collection.get_attribute_list('dest_path')
                               )
-        self.assertEqual(len(handler.file_collection), 5)
-
+        self.assertEqual(len(handler.file_collection), 7)
```
ocehugo: why the 7 hardcoding? If it's because this class is only executed for a single test, them grab the number of file_collection as args.

ocehugo: I also think the test_good_manifest is doing too much stuff - maybe some function refactoring !?

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -54,19 +54,32 @@ def test_good_manifest(self, mock_webfeatureservice):
...
+
+        # check new product files
```
ocehugo: this part can be a test function

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -54,19 +54,32 @@ def test_good_manifest(self, mock_webfeatureservice):
...
         published_files = handler.file_collection.filter_by_attribute_id('publish_type',
                                                                          PipelineFilePublishType.HARVEST_UPLOAD)
-        self.assertEqual(len(published_files), 3)
+        self.assertEqual(len(published_files), len(expected_new_products))
         for f in published_files:
             self.assertTrue(f.is_harvested and f.is_stored)
+        published_products = {get_product_type(f.name) for f in published_files}
+        self.assertSetEqual(published_products, expected_new_products)
 
+        # check deletion of previous versions
```
ocehugo: this part can also be a separate one

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -54,33 +54,44 @@ def test_good_manifest(self, mock_webfeatureservice):
...
                               handler.input_file_collection.get_attribute_list('dest_path')
                               )
-        self.assertEqual(len(handler.file_collection), 7)
 
-        # check new product files
         expected_new_products = {'TEMP-aggregated-timeseries',
                                  'PSAL-aggregated-timeseries',
                                  'CHLF-aggregated-timeseries',
                                  'hourly-timeseries',
                                  'hourly-timeseries-including-non-QC'
```
ocehugo: Again, this commit is inflating the amount of assertions in a single test. IMO if a test require a lot of assertions, them split the assertions into smaller testcalls to improve composability.

aodndata/moorings/products_handler.py
```diff
@@ -206,6 +206,16 @@ def _get_old_product_files(self):
                                                                       product_url=product_url)
             )
 
+    def _handle_errors(self, errors):
```
ocehugo: the function name is not the intent - you are just reporting and adding files to a list.
I would rename to "check_excluded_files" or alike

aodndata/moorings/products_handler.py
```diff
@@ -109,6 +110,7 @@ def __init__(self, *args, **kwargs):
         self.allowed_extensions = ['.json_manifest', '.nc', '.zip']
         self.product_site_code = None
         self.product_variables = None
+        self.products_to_create = VALID_PRODUCTS
```
ocehugo: I would set this along FILE_INDEX_LAYER since it is not owned by HandlerBase.

aodndata/moorings/products_handler.py
```diff
@@ -126,6 +128,14 @@ def _read_manifest(self):
             raise InvalidFileContentError(
                 "manifest file '{self.input_file}' missing information (site_code, variables)".format(self=self)
             )
+        if 'products' in manifest:
```
ocehugo: This if and the try statement above could be refactored into a validation function ```validate_manifest(manifest)```.

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -5,27 +5,34 @@
...
 from aodncore.pipeline.storage import get_storage_broker
 from aodncore.testlib import HandlerTestCase, make_test_file
 
 from aodndata.moorings.products_handler import MooringsProductsHandler, MooringsProductClassifier, get_product_type
 
 TEST_ROOT = os.path.dirname(__file__)
 GOOD_MANIFEST = os.path.join(TEST_ROOT, 'test_product.json_manifest')
+AGGREGATED_ONLY_MANIFEST = os.path.join(TEST_ROOT, 'test_product_aggregated.json_manifest')
+BAD_VAR_MANIFEST = os.path.join(TEST_ROOT, 'test_product_bad_var.json_manifest')
 PRODUCT_FILE = os.path.join(
```
ocehugo: wouldn't file name parameters be best to be arguments in a test function!?

test_aodndata/moorings/test_mooringsProductsHandler.py
```diff
@@ -5,27 +5,34 @@
...
     TEST_ROOT,
     'IMOS_ANMN-NRS_TZ_20181213_NRSROT_FV01_TEMP-aggregated-timeseries_END-20190523_C-20191218.nc'
 )
 
 GETFEATURE_FILE = os.path.join(TEST_ROOT, 'getFeature.json')
 GETFEATURE_OLD_PRODUCTS_FILE = os.path.join(TEST_ROOT, 'getFeature_old_products.json')
+GETFEATURE_EMPTY_FILE = os.path.join(TEST_ROOT, 'getFeature_empty.json')
```
ocehugo: I think those opens should be test functions!?

test_aodndata/moorings/test_mooringsProductsHandler.py
ocehugo: Ahh nevermind - you are loading cases here - better than put a little comment above.

test_aodndata/moorings/test_mooringsProductsHandler.py
ocehugo: Gotcha - also, put a comment alike "loading the cases".

aodndata/moorings/products_handler.py
mhidas: Done

aodndata/moorings/products_handler.py
mhidas: This is only relevant to the hourly product. I've removed the class attribute, hard-coded the values in the `_make_hourly_timeseries` method, and added a comment to explain it.

aodndata/moorings/products_handler.py
mhidas: renamed to `_log_excluded_files`

aodndata/moorings/products_handler.py
mhidas: I've updated the docstring and added a comment to explain this better.

test_aodndata/moorings/test_mooringsProductsHandler.py
mhidas: comments added

test_aodndata/moorings/test_mooringsProductsHandler.py
mhidas: This handler is doing quite a lot of things, so I make a lot of assertions to check different aspects, but they all refer to the
_same test case_ (i.e. input file). Possibly it could all be refactored to run the test case just once, then make different assertions in separate methods independently, but at this point I don't think it's worth the effort.

aodndata/moorings/products_handler.py
mhidas: DOne

aodndata/moorings/products_handler.py
mhidas: It could. For now I don't think it's worth it. "products" is optional, and the other two keys are already being validated above.

mhidas:
Thanks @ocehugo . I have responded to your comments and made some changes. Please merge if you're happy with it.

aodndata/moorings/products_handler.py
ocehugo: OK.

I still think the file_collection.add(old_file) is a bit anti-pattern because it leads the reader to think you are adding a file to a collection but you are in-fact scheduling the file to be removed later - so I would put more intent via a wrapper:

```
def sched_for_removal(self, file: str, **kwargs):
pfile = PipelineFile(file,**kwargs)
pfile.publish_type = ...
self.file_collection.add(pfile)
```
Also, the publish_type should be settable as kwargs too

aodndata/moorings/products_handler.py
ocehugo: I understand, but to know what this will cause down the road was a bit convoluted since the aodntools function that uses it got several parameters, several defaults, and is 500lines or more. 
For example, if you use the actual names inside the aodntools function (hourly-non-qc-whatver) and do string comparison instead you make a safe-closure. If someone add another case here (e.g. [0,1]) the [0,1,2] collection is overwritten.

aodndata/moorings/products_handler.py
mhidas: Yes, I _am_ adding a file to a collection. `self.file_collection` is the set of files the handler is operating on. Removing a file is one of the valid operations it can perform!

Anyway, I understand what you mean, but I think the comments I've added (plus an understanding of how `HandlerBase` works) should make it clear enough.

aodndata/moorings/products_handler.py
mhidas: Good point, but fixing that will require an API change to the function in aodntools. I've added this as a future enhancement (https://github.com/aodn/python-aodntools/issues/100)

mhidas:
Thanks for the comments @ocehugo . I might address some of them later, but I'd like to merge this now.

PR #212, opened 2020-03-16
Add gridded timeseries product to MooringsProductsHandler

mhidas:
By default the handler will now produce the aggregated, hourly, and gridded product from all non-velocity FV01 files at a site.

Note that as the gridded product takes an hourly file as input, even if only the gridded is requested, the handler will re-create (and publish) the hourly product too.


aodndata/moorings/products_handler.py
```diff
@@ -274,6 +275,39 @@ def _make_hourly_timeseries(self):
...
+        hourly_file = hourly_files[0]
+        self.logger.info("Creating gridded product from '{hourly_file.name}'".format(hourly_file=hourly_file))
+
+        # create a local symlink in temp_dir which includes the final S3 id (dest_path) so that
+        # it is recorded in the gridded file's metadata
+        hourly_file.dest_path = self.dest_path(hourly_file.local_path)
+        hourly_temp_path = os.path.join(self.temp_dir, hourly_file.dest_path)
+        os.makedirs(os.path.dirname(hourly_temp_path))
+        os.symlink(hourly_file.local_path, hourly_temp_path)
+
```
ocehugo: is this shell-like hack really necessary !? I can fanthom why only with this codebase - apparently a plumbing to an aodncore missing functionality.

aodndata/moorings/products_handler.py
```diff
@@ -310,9 +344,15 @@ def preprocess(self):
 
         if 'aggregated' in self.products_to_create:
             self._make_aggregated_timeseries()
```
ocehugo: Up to you, but it's I was expecting a new class of MooringProductHandler to be created for each type of product - after all they are all different classes. 
The pros is that this would clearly avoid any possible side effects and reduce logic handling of nested dependencies.

aodndata/moorings/products_handler.py
```diff
@@ -310,9 +344,15 @@ def preprocess(self):
...
-        if 'hourly' in self.products_to_create:
+
```
ocehugo: Maybe a good idea is to explicitly set the dependencies instead of using top-bottom separated if:
```
gridded_product_mlist = [self.make_aggregated_timeseries, self.make_hourly_timeseries, self.make_gridded_timeseries ]
hourly_product_mlist = [...]
...
if 'gridded' in self.products_to_create:
clist = gridded_product_mlist
elif  ...
...
for create_product in clist:
     self.logger.info(...)
     create_product()

File: test_aodndata/moorings/getFeature.json [hunk 4]
```diff
@@ -6,7 +6,7 @@
             "id": "moorings_all_map.fid--44e8da32_16d0014d48c_7b17",
             "geometry": null,
             "properties": {
-                "url":"IMOS/ANMN/NRS/NRSROT/Temperature/IMOS_ANMN-NRS_TZ_20180406T080000Z_NRSROT_FV01_NRSROT-1804-SBE39-27_END-20180817T023000Z_C-20180820T010304Z.nc",
+                "url":"IMOS/ANMN/NRS/NRSROT/Temperature/IMOS_ANMN-NRS_TZ_20180406T080000Z_NRSROT_FV01_NRSROT-1804-SBE39-27_END-20180413T064000Z_C-20200313T000000Z.nc",
```
ocehugo: This things lead to a very high maintenance burden - just mock that up somehow.

ocehugo: Same as for all others below

aodndata/moorings/products_handler.py
lwgordonimos: aodncore has nothing to do with *how* NetCDF files are generated. 

This appears to be parameterising this function in aodntools according to the expectations of *that* function:

https://github.com/aodn/python-aodntools/blob/master/aodntools/timeseries_products/gridded_timeseries.py#L139

mhidas: That's right. As stated in the comment above (line 291), this is done so that the gridded product can include its source file's S3 id (making it directly accessible).

The alternative would be to add a separate argument to the gridding function, specifying this id. This would be ok in this case, but more complicated for the other products, which have a list of input files (often hundreds...), so ensuring the two lists are in sync would be a mess. So I'd rather just keep this solution.

aodndata/moorings/products_handler.py
mhidas: I think it makes sense to do it have the products generated by the same process, because
1. whenever there's new data for a site, we'll want to re-generate _all_ the products for that site;
1. multiple products are generated from the same set of input files; and
1. the gridded product takes the hourly as an input.

test_aodndata/moorings/getFeature.json
mhidas: This _is_ the mocked response from the WFS server. There's nothing to maintain. The files themselves are included in the repo.

mhidas:
This is ready to merge now!

mhidas:
Thanks @ocehugo !

PR #213, opened 2020-04-08
feat(fix): fix py3.5 broken by adding constraints

ocehugo:


setup.py
```diff
@@ -58,6 +58,10 @@
     'cc-plugin-imos>=1.3.0',
     'fiona>=1.8.8',
     'matplotlib>=3.0.3',
+    'numpy>=1.13.0',
```
lwgordonimos, line 61: These dependencies are not required by *this* package, so don't need to be in here. The are either "dependencies of dependencies" or provided by the dependencies anyway. We've found through trial and error that keeping the requirements minimal here saves a lot of conflicts and "fixup" PRs later.

ocehugo: OK - I include that because of my pip experience with circular dependencies and thought that failing towards the leaf was best.

On the funny side, this is a fixup PR because of that same reason! Maybe we need a jenkins job just for dependencies check then?

setup.py
```diff
@@ -58,6 +58,8 @@
     'cc-plugin-imos>=1.3.0',
     'fiona>=1.8.8',
     'matplotlib>=3.0.3',
+    'pandas>=0.24.2,<0.25.0',
```
lwgordonimos, line 61: These lines are unnecessary. We just need to add constraints to the *environment* (as you've done in requirements.txt/constraints.txt), not add dependencies to the *package*. The reason is that it already gets these dependencies via the aodntools package. Having these here as well just means it will require synchronising updates to these versions with the aodntools package, which is a pain when it comes to updating/deploying them if they get out of sync.

PR #214, opened 2020-05-06
Moorings Products: add velocity

mhidas:
A bit more refactoring could be done (probably later), but all the required functionality is there.


aodndata/moorings/products_handler.py
```diff
@@ -139,6 +141,20 @@ def _read_manifest(self):
                 )
             self.products_to_create = set(manifest['products'])
 
+        # Even if only the gridded product is explicitly requested, we need to re-generate and publish the hourly too,
+        # as it is the input file for the gridded.
+        if 'gridded' in self.products_to_create:
+            self.products_to_create.add('hourly')
+
+    @property
```
ocehugo, line 149: I would change the name to something like "input/output options " or alike to raise awareness that it's just IO options.

aodndata/moorings/products_handler.py
```diff
@@ -224,6 +240,12 @@ def _log_excluded_files(self, errors):
                 else:
                     self.excluded_files[f].update(e)
 
+    def _add_to_collection(self, product_url):
+        """Add a new product file to the file_collection to be harvested and uploaded."""
+        product_file = PipelineFile(product_url, file_update_callback=self._file_update_callback)
```
ocehugo, line 245: Lost the account of the number of times I see this pattern of setting a field after initialization. We should allow setting the publish_type at PipelineFile at init time too : ```PipelineFile(...,publish_type= ...HARVEST_UPLOAD)```. I've to remember to raise this in aodncore - any thoughts?

aodndata/moorings/products_handler.py
```diff
@@ -236,44 +258,64 @@ def _make_aggregated_timeseries(self):
...
             self._log_excluded_files(errors)
+            self._add_to_collection(product_url)
+            self._cleanup_previous_version(os.path.basename(product_url))
+
+    def _make_velocity_aggregated_timeseries(self):
+        """Generate the velocity aggregated timeseries product and add to file_collection."""
+
+        # Filter input list to just the velocity files, all of which will have the variable 'VCUR'
+        input_list = [f for f, f_vars in self.input_file_variables.items()
+                      if 'VCUR' in f_vars
```
ocehugo, line 272: I would change the comment to highlight why VCUR is used as a matching filter instead.

mhidas: Would this be enough to make it clear? (If not, then I'm not sure what you mean)
```python
# Filter input list to just the velocity files, all of which will have
# the variable 'VCUR' (= "northward_sea_water_velocity")
```

aodndata/moorings/products_handler.py
mhidas: They're not all "input/output options" (the "*_url_prefix" args end up written to variable attributes in the output files).
The one thing they all have in common is that all the product-generating functions accept these args, so I thought this name was fairly descriptive.

aodndata/moorings/products_handler.py
mhidas: I see your point, but that's nothing to do with this PR. It's an aodncore issue

mhidas:
Thanks @ocehugo . I have responded to your comments. Happy to update that code comment about filtering the input list using "VCUR". Let me know if that will work.

mhidas:
@ocehugo I have updated the filtering of input files for the velocity products, and hopefully made the comment a bit more informative. I also did a minor refactor while I was at it, and included a quick fix for a bug I just found. See last three commits.

Unless you find any new issues, please merge this so I can test & deploy!

aodndata/moorings/products_handler.py
ocehugo: It was more like a warning because of the simple matching filter - I imagine you are using VCUR as a placeholder for ``` if 'UCUR' in ... and 'VCUR' in ... and ... other possible conditions matched by files containing VCUR variables ```

aodndata/moorings/products_handler.py
```diff
@@ -246,14 +246,19 @@ def _add_to_collection(self, product_url):
         product_file.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
         self.file_collection.add(product_file)
 
+    def _input_list_for_variables(self, *variables):
+        """Return a list of input files containing any of the given variables"""
+        input_list = [f for f, f_vars in self.input_file_variables.items()
+                      if any(v in f_vars for v in variables)
```
ocehugo, line 252: would it be `all` instead of ```any``` to avoid the 0.0001% chance of a file without UCUR/VCUR but with VCUR/UCUR to be included and raise a runtime error down the line?

ocehugo: @mhidas - just confirm the above and free to merge

mhidas: Ideally, yes, all velocity files _should_ have UCUR _and_ VCUR (and also WCUR), and in fact pretty much all of them do. However, if there happen to be any files with only one and not the other, we want to know about it, not just silently ignore it by not selecting it here.
Also, the product generating code [checks](https://github.com/aodn/python-aodntools/blob/ae8b06eff10a2143f0f7766a82844c433af4634c/aodntools/timeseries_products/velocity_aggregated_timeseries.py#L20) each input file before processing and excludes files that don't meet the minimum requirements, so these won't lead to runtime errors. Instead they are reported on output and also listed in a global attribute of the product file.

PR #215, opened 2020-05-20
Update pattern for image zip files accepted by ABOS handler

mhidas:


mhidas:
@ocehugo This should be a quick one.

mhidas:
Thanks @ocehugo !

PR #216, opened 2020-05-20
Make excluded files and error messages easier to grep from pipeline log

mhidas:


mhidas:
@ocehugo Another very quick one, if you wouldn't mind... 
This is just to make it a bit easier to pull out the list of rejected files and their errors from the pipeline log (see https://github.com/aodn/PO-Backlog/issues/750#issuecomment-627742103)

PR #217, opened 2020-05-21
Update to source AODN dependencies from custom PyPI repo

lwgordonimos:
We are now publishing our Python packages to custom PyPI repositories, so we can do away with the custom download code, and use native pip commands for sourcing the packages.


lwgordonimos:
@mhidas @lbesnard if you get a spare 5 minutes, I'd value your input with the doco and virtualenv setup side of things. It should be a pretty transparent change, and I've tried to make the requirements side of thing as generic as possible between Travis/Jenkins/local environments.

scripts/setup_virtualenv.sh
```diff
@@ -2,14 +2,15 @@
...
+STAGE=${STAGE:-production}
 VIRTUALENV_DIR=${VIRTUALENV_DIR:-python-aodndata-virtualenv}
 
 HERE=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 echo "##### Creating/updating virtual environment #####"
 virtualenv -p /usr/bin/python3 ${VIRTUALENV_DIR}
 source ${VIRTUALENV_DIR}/bin/activate
-${HERE}/install_dependencies.sh
+pip install -r stage_requirements.txt
```
lbesnard, line 13: what do you think about using /usr/bin/pip instead, in order to make sure there is no confusion in case someone is already on a different virtual env?

lwgordonimos: This is after the virtual has just been activated (with source), so should only ever be the **virtualenv** pip version. The only thing here which would be exposed to being called from an existing virtualenv would be the `virtualenv` command itself, but that's unlikely to matter.

If you have already made a virtualenv, you don't need this script at all, you just just run pip yourself, e.g.

```bash
$ export STAGE=rc
$ pip install -r stage_requirements.txt

# Or even this if you just want to run prod dependencies without the environment variable
$ pip install -r test_requirements.txt
```

This script is just a very basic one to automate a single function of making a virtualenv with the package installed in it, but there's nothing at all stopping you installing the requirements directly. In fact, that's what I'm leaning towards trying to m

lbesnard: I was more talking in my case for instance, where my home pip is different from my system wide pip

lwgordonimos: Yeah it shouldn't matter at all. The activate puts the virtualenv on the front of the PATH, so all your other pips aren't involved here at all:

```bash
$ source ${VIRTUALENV_DIR}/bin/activate
$ which pip
/home/me/src/python-aodndata/python-aodndata-virtualenv/bin/pip
```

PR #223, opened 2020-08-21
Update .travis.yml

ocehugo:


lwgordonimos:
FYI I found this issue the other day, and I'm pretty sure it's because of the new version of the Python NetCDF library: https://pypi.org/project/netCDF4/1.5.4/#files

It's not providing pre-compiled binaries suitable for Ubuntu Xenial unfortunately

So it's probably better to tell Travis to use Bionic, then it wouldn't need to muck around trying to compile things:

https://docs.travis-ci.com/user/reference/bionic/#using-ubuntu-1804-bionic-beaver

`dist: bionic`

ocehugo:
A-ha - so used to have everything compiled that I almost forgot these distribution problems.

Yeap, I think pinning to 1.5.3 is best since this is the prod/rc version.

lwgordonimos:
Yeah I'd lean towards pinning, as the most conservative option. I don't think there's any requirement to rush onto netCDF 1.5.4.

PR #224, opened 2020-08-21
feat(bugfix): include req. sys libs for netcdf4

ocehugo:
~~this add required libs to the docker image, so both travis and jenkins can bootstrap the python-netcdf4 dependency  py3.5) for the tests.~~
see commit msg

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,3 +1,4 @@
 cftime<1.1.1;python_version=='3.5'
 pandas<0.25.0;python_version=='3.5'
 xarray<0.14.0;python_version=='3.5'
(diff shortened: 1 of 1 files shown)


ocehugo:
I think I may have ressurected some skeletons here when fixing the py35 test runs

lwgordonimos:
I think the pinning option is still the best!

Have a look here: https://github.com/aodn/python-aodncore/pull/213

Pinning netCDF makes everything work again on Xenial, just like it did before July 20 ;)

ocehugo:
OBEY!

PR #230, opened 2020-09-22
Aatams harvest only zipfile

ocehugo:
see commits msg


aodndata/aatams/aatams_sattag.py
```diff
@@ -172,8 +220,8 @@ def process(self):
 AATAMS_SATTAG_QC_DM_OPTS = {
     "allowed_extensions": [".zip", ".csv"],
     "allowed_regexes": ["^.+_dm\\.(zip|csv)$"],
-    "allowed_archive_path_regexes": [AATAMS_SATTAG_QC_DM_BASE + ".+\\.(zip|csv)"],
-    "allowed_dest_path_regexes": [AATAMS_SATTAG_QC_DM_BASE + ".+\\.(zip|csv)"],
+    "allowed_archive_path_regexes": [AATAMS_SATTAG_QC_DM_BASE + ".+\\.(zip|csv)$"],
+    "allowed_dest_path_regexes": [AATAMS_SATTAG_QC_DM_BASE + ".+\\.(zip|csv)$"],
```
evacougnon, line 224: if we only store zip files on S3, could we remove `|csv` in the `allowed_dest_path_regexes` and `allowed_archive_path_regexes`?

What is the difference between these 2 paths?

aodndata/aatams/aatams_sattag.py
```diff
@@ -185,8 +233,8 @@ def process(self):
 AATAMS_SATTAG_QC_NRT_OPTS = {
     "allowed_extensions": [".zip", ".csv"],
     "allowed_regexes": ["^.+_nrt\\.(zip|csv)$"],
-    "allowed_archive_path_regexes": [AATAMS_SATTAG_QC_NRT_BASE + ".+\\.(zip|csv)"],
-    "allowed_dest_path_regexes": [AATAMS_SATTAG_QC_NRT_BASE + ".+\\.(zip|csv)"],
+    "allowed_archive_path_regexes": [AATAMS_SATTAG_QC_NRT_BASE + ".+\\.(zip|csv)$"],
+    "allowed_dest_path_regexes": [AATAMS_SATTAG_QC_NRT_BASE + ".+\\.(zip|csv)$"],
```
evacougnon, line 237: same than just above

evacougnon (commented):
@ocehugo, please have a look at my comments, I had a few questions.

aodndata/aatams/aatams_sattag.py
```diff
@@ -9,19 +9,19 @@
...
 NRT_TIMESTAMP_COMPARISON_MSG = (
-    "NRT update requested: Comparing timestamps in the metadata files {0} and {1}"
+    "NRT update requested: Comparing timestamps of metadata file in {0} and {1}"
 )
 NRT_TIMESTAMP_DIFFERS_MSG = "Incoming file {incoming_file} containing {within_file} contains older entries than current NRT state from {archival_file}"
 NRT_ZIPFILE_WITH_DIFF_CAMPAIGNS = (
     "Incoming file {zip_file} contains more than 1 campaign: {campaigns}"
 )
-
+NRT_NO_CAMPAIGN = "Couldn't detect campaign within incoming file {incoming_file}."
```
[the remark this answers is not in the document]
ocehugo: Good point. This msg was for when not a single csv in the received zip contains the "campaign" in the csv filename (e.g. ct155). This actually doesn't belong here, but within the schema validator. This way, we avoid handling this case by raising this problem earlier in the call stack.

aodndata/aatams/aatams_sattag.py
```diff
@@ -41,8 +41,13 @@ class AatamsSattagHandler(HandlerBase):
 
     @staticmethod
     def get_campaigns(file_list):
-        """Return all campaign ids from a PipelineFileCollection."""
-        return [x.name.split("_")[1] for x in file_list]
+        """Return the campaign id from AATAMS filenames present in
```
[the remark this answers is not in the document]
ocehugo: ATM, yes. this function operates only on "self.file_collection", which are the received files.

aodndata/aatams/aatams_sattag.py
```diff
@@ -97,42 +146,34 @@ def schedule_file_removal(self, remote_pipeline_file):
...
-            self.state_query.download(
-                RemotePipelineFileCollection(old_metadata_file), self.temp_dir
-            )
-        except AttributeError:
-            self.state_query._storage_broker.download(
-                RemotePipelineFileCollection(old_metadata_file), (self.temp_dir)
-            )
 
         if not self.is_nrt_update_required(old_metadata_file.local_path):
             raise InvalidFileContentError(
```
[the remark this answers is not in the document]
ocehugo: Yeap, good catch. I didn't add a test for this particular error msg, although I test another specific error msg.

This is now happening because the metadata file is not stored on s3 anymore. Hence, we lost the ```dest_path``` property. 

Fix is quick - just need to point to the old_zip_file it's comparing against and add a specific test for this msg.

aodndata/aatams/aatams_sattag.py
ocehugo: Yes, you and me agree, but python-aodncore doesnt! python-aodncore considers these attributes for all files within a file_collection, and such all csvs within the zip.

aodndata/aatams/aatams_sattag.py
ocehugo: as above

aodndata/aatams/aatams_sattag.py
lwgordonimos: Remember `dest_path` acts as the key for both S3 AND the database, so it pays not to think of it purely in terms of S3 uploads.

In the most general use case, there is a row in the database AND an object on S3, but for any harvested files, it's probably more useful to think of it more like this:

* `dest_path` determines the `file_url/url` in the database and there is usually a corresponding file on S3
* `allowed_dest_path_regexes` is a generic validation of what the value of this "key" is allowed to be
* for UPLOAD_ONLY and HARVEST_ONLY files, it's getting to be a bit of a niche requirement, so it's not worth overcomplicating things for the generic case.
* if it shouldn't be uploaded at all, the handler should simply not flag CSV files with an UPLOAD publish type... this mechanism doesn't exist to duplicate that kind of general handler logic
* `allowed_dest_path_regexes` is not really meant to be super pedantic about extensions, it's more for things like preventing a bug in one pipeline from generating prefixes outside the pipeline's "owned prefix"

This is just an additional overall safety thing to make sure that pipelines only operate on their own prefixes. For example, AATAMS should never be interacting with a `dest_path` that doesn't start with `IMOS/AATAMS`. This applies equally to the database fields as it does to the S3 prefixes, hence the single `dest_path` validation mechanism which applies to both harvested URLs AND S3 objects.

> What is the difference between these 2 paths?

"Archiving" means the input file it gets copied off to 10-nec-hob and stored at TPAC, but is *not* published to S3. `archive_path` determines how this is stored in the archive, because during development of Pipeline 2 there was a requirement that this should in some cases be different from dest_path.

ocehugo: I tried to be short but missed some things in the process...

Two things:
1. the archival regex can't match the csv bit here. this is a mistake since we read zips, harvests csvs, but only s3-store/archive the zip.

  2. However, we don't actually need any of the file matchings in these regexes, because we do this explicitly for every item in the collection and the file_object itself. Including them here is just an historical artifact and me trying to match new requirements (store the zip in s3). All because now we got a mixed bag file_collection with different files/publishing properties (zip+csvs).

My take is that this "dest_path/archive_path" is ambiguous in behaviour as is the name itself,since  ```dest_path/archive_path``` is a file suffix. Funny enough, it is also named as other things too (like file_url, which is also not a url).

ocehugo:
@evacougnon , comments addressed here https://github.com/aodn/python-aodndata/pull/230/commits/0c9d28c68de77b7a6dc8101f53db75d208fabc1a

evacougnon (approved):

PR #234, opened 2020-11-10
Aatams allow empty dive files

ocehugo:
This allows the AATAMS SATTAG pipeline to bypass the validation of empty dive.csv files. I found best to ignore the empty file than to force creation of an meaningless one.

The behaviour proposed is to mark the CSV file as a UNKNOWN FileType, so it is ignored by the harvester. I hope this makes the harvester modifications to a minimum (just accept that dive files may be missing).


aodndata/aatams/aatams_sattag.py
```diff
@@ -167,7 +170,20 @@ def preprocess(self):
         except ValueError as e:
             raise InvalidFileFormatError(''.join(e.args))
 
+        dive_file = self.get_file(self.file_collection, 'dive')
+        if dive_file:
+            is_dive_csv_empty = not dive_file.file_type.validator(dive_file.local_path)
+            if is_dive_csv_empty:
+                # skip harvesting, filetype/content/cross validations
```
mhidas, line 177: I don't know much about how this particular handler works, but in general a file you don't want to validate would have
```python
dive_file.check_type = PipelineFileCheckType.NO_ACTION
```

lwgordonimos: Yes, this is definitely getting lost in the library internals unnecessarily.

It's a fairly declarative interface, where you can make sure a file is ignored by just setting the attributes:

```python
dive_file.publish_type = PipelineFilePublishType.NO_ACTION
dive_file.check_type = PipelineFileCheckType.NO_ACTION
```

It's largely because the documentation for this is pretty inadequate currently, but these high level core concepts should be very prominently featured in the Quick Start doco here, but https://aodn.github.io/python-aodncore/, but it's a bit buried down in the module documentation: https://aodn.github.io/python-aodncore/aodncore.pipeline.html#aodncore.pipeline.common.PipelineFileCheckType

ocehugo: Yeap, fixed. I went to where the stack trace was pointing the finger. Since the file_type.validator method - and myself - don't know all the higher level uses of it, I just used what was in the scope. I also blame pdb which is very bad at method instrospection.

mhidas:
@ocehugo @lbesnard FYI, I have built and tested the _next_ version of aodndata (1.3.20). That version also includes #235, which is just decommissioning some old "ABOS" code, so not urgent. But whenever this AATAMS change is good to go, please promote 1.3.20 to prod, not 1.3.19. Cheers.

PR #236, opened 2020-12-04
Fix nrt campaign handling

ocehugo:
This provides more strict handling of the AATAMS campaign ids. Before we were only checking campaign id at zip filename, and this was a very simple string split extraction/check. This created a bug on NRT updates when checking campaigns.

Now, the campaign id check is updated to the word between underscores, but more importantly, it is reinforced at every csv filename. 

See commit msgs for more details.


ocehugo:
@lbesnard, @mhidas - this is now ready for review.

Sorry to poke you early, but the previous PR solution was breaking a case that I forgot to handle.

test_aodndata/aatams/test_aatams_sattag.py
```diff
@@ -18,13 +18,15 @@
...
+DM_SINGLE_CSV = os.path.join(TEST_ROOT, "metadata_ct111_dm.csv")
 
 NRT_FIRST_ZIP = os.path.join(TEST_ROOT, "first_nrt_batch", "ct155_nrt.zip")
 NRT_SECOND_ZIP = os.path.join(TEST_ROOT, "second_nrt_batch", "ct155_nrt.zip")
 NRT_NEW_CAMPAIGN = os.path.join(TEST_ROOT, "new_ct156_nrt.zip")
-NRT_SINGLE_CSV = os.path.join(TEST_ROOT, "metadata_nrt.csv")
+NRT_SINGLE_CSV = os.path.join(TEST_ROOT, "metadata_ct111_nrt.csv")
 NRT_EMPTY_DIVE_ZIP = os.path.join(TEST_ROOT, "emptydive_ct999_nrt.zip")
+NRT_MIXED_CAMPAIGN = os.path.join(TEST_ROOT, "mixcampaign_ct555_nrt.zip")
+NRT_CSV_CAMPAIGN_MISMATCH = os.path.join(TEST_ROOT, "csvcampaignmismatch_ct999_nrt.zip")
```
mhidas, line 29: Just FYI, you can also create test zip files on the fly with [`aodncore.testlib.make_zip`](https://github.com/aodn/python-aodncore/blob/master/aodncore/testlib/testutil.py#L187)

mhidas (commented):
Looks mostly ok to me. I don't know much about these campaigns and the logic around their codes, but I couldn't figure our your "campaignmismatch" test case - it looks to me like the zip file name and all the files within have 'ct999'. I also don't see where in the code you check for this mismatch?

test_aodndata/aatams/test_aatams_sattag.py
```diff
@@ -143,9 +144,32 @@ def test_no_clobber_if_different_campaign(self):
...
+        self.run_handler(NRT_SECOND_ZIP)
+        self.run_handler(NRT_NEW_CAMPAIGN)
+        self.run_handler(NRT_NEW_CAMPAIGN)
+        self.run_handler(NRT_EMPTY_DIVE_ZIP)
         handler = self.run_handler(NRT_EMPTY_DIVE_ZIP)
-
+        handler_dest_path = handler.dest_path_function("")
+        expected = {os.path.basename(x) for x in [NRT_FIRST_ZIP, NRT_NEW_CAMPAIGN, NRT_EMPTY_DIVE_ZIP]}
+        result = {os.path.basename(x.dest_path) for x in handler.state_query.query_storage(handler_dest_path)}
+        assert result == expected
```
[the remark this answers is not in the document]
ocehugo: I'm biased towards the terser one because of my python checkers, but yeah, the more verbose call is the defacto standard.

ocehugo:
> Looks mostly ok to me. I don't know much about these campaigns and the logic around their codes, but I couldn't figure our your "campaignmismatch" test case - it looks to me like the zip file name and all the files within have 'ct999'. I also don't see where in the code you check for this mismatch?

The "campaignmismatch" file is to test content instead of filename strings. This is part of cross validation (see `cross_set_validation` in `aatams_sattag_schema.py`). It was implemented since the start, is compulsory, but a test was lacking. I included it here because I realise the lack of a specific test when making these campaign filename handlings.

For example, If you look into the `cid` column of the "campaignmismatch" file, there will be a mixed bag of "ct999" and others. This was cared for already, but now we have a specific test.

test_aodndata/aatams/test_aatams_sattag.py
ocehugo: Yeah, I may refactor into csv-folders if we ever need more zips.

ocehugo: I just quickly tried to move towards using makezip for all current zip files, but this would incur in way more code, more file creation,  and less re-use. The problem lies with `TemporaryDirectory` removing the temporary folder when the scope finishes. Hence, I would need to copy/paste the Temporary scope into all tests, build zips in every test, rebuild several zips in the tests that use more than one zip file, etc. Here they are a single line definition.

The proper solution would involve a TemporaryDirectory scope that is bounded to the test runtime - a thing pytest already got.


I'm leaving as is.

PR #248, opened 2021-01-12
Soop xbt nrt - various fixes

lbesnard:


lbesnard:
@lwgordonimos would you have any idea why the 3.6 build is failing the unittest?

lwgordonimos:
After some IDE testing, I think we've found the issue.

This test also calls the `ship_callsign` function: https://github.com/aodn/python-aodndata/blob/master/test_aodndata/soop/test_ship_callsign.py#L17, and the mocked output does *not* contain your new ships.

Because this has the lru_cache decorator it's caching the first list of ships, which is why when these soop_xbt_nrt tests are run, the ship is not found.

This would be less quirky I think if you could mock the `ship_callsign_list` function itself, so there's never any cache related issues in tests.

lbesnard:
thanks @lwgordonimos, I should have run all the unittests of the repo. I opted for clearing the cache of that function. way easier/faster to implement

bpasquer (approved):

PR #249, opened 2021-01-14
Updates to make use of aodncore API improvements

mhidas:
https://github.com/aodn/python-aodncore/pull/209 introduces a couple of API improvements for:
* handlers explicitly creating and adding `PipelineFile` objects to their `file_collection`; and
* handlers downloading remote files
The old functionality is still there, but now issues deprecation warnings. 

This PR is to update all handlers to use the new API, so the old methods can be removed in a future aodncore update.

### DIFF
--- a/aodndata/moorings/handlers.py
+++ b/aodndata/moorings/handlers.py
@@ -59,9 +59,8 @@ class MooringsHandler(HandlerBase):
 
             self.logger.info("Burst-processing {f.name}".format(f=f))
             product_path = create_burst_average_netcdf(f.src_path, self.products_dir)
-            product_file = PipelineFile(product_path, file_update_callback=self._file_update_callback)
-            product_file.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
-            self.file_collection.add(product_file)
+            product_file = PipelineFile(product_path, publish_type=PipelineFilePublishType.HARVEST_UPLOAD)
+            self.add_pipelinefile(product_file)
 
     def postprocess(self):
         """Set error_cleanup_regexes so that if the same file was uploaded previously and failed, it can now be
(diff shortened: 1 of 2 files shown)


test_aodndata/soop/test_soop_trv.py
```diff
@@ -79,7 +79,7 @@ def test_delete_from_wfs(self, mock_webfeatureservice):
 
         broker = WfsBroker(self.config.pipeline_config['global']['wfs_url'])
 
-        files_for_layer = broker.query_urls_for_layer('soop_trv_duplicate_url')
+        files_for_layer = broker.query_files('soop_trv_duplicate_url')
         self.assertEqual(files_for_layer[0],
                          'IMOS/SOOP/SOOP-TRV/VMQ9273_Solander/By_Cruise/Cruise_START-20181205T035932Z_END-20181206T045722Z/temperature/IMOS_SOOP-TRV_T_20181205T035932Z_VMQ9273_FV01_END-20181206T045722Z.nc')
```
mhidas, line 84: @lwgordonimos  Should this test be creating a `WfsBroker` instead of using the new `state_query` api?
e.g.
```suggestion
        handler = self.handler_class(GOOD_NC)
        files_for_layer = handler.state_query.query_wfs_files('soop_trv_duplicate_url')
        self.assertEqual(files_for_layer[0].dest_path,
                         'IMOS/SOOP/SOOP-TRV/VMQ9273_Solander/By_Cruise/Cruise_START-20181205T035932Z_END-20181206T045722Z/temperature/IMOS_SOOP-TRV_T_20181205T035932Z_VMQ9273_FV01_END-20181206T045722Z.nc')
```

In fact, this is already implicitly tested by `test_good_netcdf` above, so this one is kind of rendundant.
@lbesnard

PR #251, opened 2021-01-17
Update test.yml

lwgordonimos:
Add fail-fast: false


mhidas:
I've added this to #249 and it seems to be doing what we want. Thanks @lwgordonimos !

PR #255, opened 2021-02-22
pipeline update for NTP wave data

ana-berger:


test_aodndata/aodn_wave_dm/test_aodn_wave_dm.py
```diff
@@ -96,6 +96,24 @@ def test_dest_path_aodn_wave_dm(self):
                              'Southern_Ocean',
                              os.path.basename(good_nc)))
 
+        good_nc = os.path.join(TEST_ROOT, 'IMOS_NTP-WAVE_TW_20200113T215850Z_TOR01_WAVERIDER_FV01_timeseries_END-20200319T015105Z.nc')
```
mhidas, line 99: FYI: This file name doesn't match the IMOS file naming convention:
`IMOS_<Facility-Code>_<Data-Code>_<Start-date>_<Platform-Code>_FV<File-Version>_<Product-Type>_<End-date>_<Creation-date>_<PARTX>.nc`

Between `<Start-date>` and `FV01`, there should only be a single field `<Platform-Code>`. You have two fields there `TOR01` and `WAVERIDER`. Which is the platform code? If you want to keep them both there, you should put a hyphen between them, instead of an underscore.

mhidas: Alternatively, (e.g. if this file were part of the moorings facility) you might think of "WAVERIDER" as the _instrument_, and put it in the `<Product-Type>` field:
`IMOS_NTP-WAVE_TW_20200113T215850Z_TOR01_FV01_WAVERIDER-timeseries_END-20200319T015105Z.nc`

aodndata/aodn_wave_dm/aodn_wave_dm.py
```diff
@@ -97,6 +112,18 @@ def dest_path_aodn_wave_dm(filepath):
                 "to deduce path".format(filename=file_basename))
         product_dir = site_name.replace('Wave Rider Buoy', '').strip().replace(' ', '_')
 
+    elif NTP_WA_WAVERIDER_TW.match(file_basename):
+        data_base_dir = os.path.join(NTP_WA_DIR, WAVERIDER_DIR, DELAYED_DIR)
+        if 'Torbay01' not in site_name:
+            raise InvalidFileContentError(
+                "file name: \"{filename}\"; global attribute site_name does not contain 'Torbay01' string to " \
+                "to deduce path".format(filename=file_basename))
```
mhidas, line 120: Duplicated word "to to deduce..." (also in lines 111/112 above)

test_aodndata/aodn_wave_dm/test_aodn_wave_dm.py
bpasquer: Yes, @mhidas the filename does not follow the IMOS convention, but it is consistent with the format adopted for the files of the National wave archive which this dataset will be part of.

PR #266, opened 2021-05-04
update DWM/SOTS pipeline to include calibration zip files

ana-berger:
@mhidas I did changes in the moorings handler and test_dwmHandler. The tests ran alright, but you might find some non-sense coding. Anyway, it is a starting for discussion and it is working, which is a good sign! This is part of [issue#755](https://github.com/aodn/PO-Backlog/issues/755)

### DIFF
--- a/aodndata/moorings/handlers.py
+++ b/aodndata/moorings/handlers.py
@@ -90,8 +90,8 @@ class DwmHandler(MooringsHandler):
         self.allowed_extensions = ['.nc', '.zip']
 
     def process(self):
-        """Handle a zip file containing images and no NetCDF files. In this case we just want to publish the zip file
-        itself, not the individual images. If we encounter a "mixed" zip file with images and netCDF files,
+        """Handle a zip file containing images and no NetCDF files or a zip file containing calibration files and no NetCDF files. In this case we just want to publish the zip file
+        itself, not the individual images, and the same for the calibration zip files. If we encounter a "mixed" zip file with images/calibration files and netCDF files,
         we're just going to give up, for now.
         """        
         images = PipelineFileCollection(f for f in self.file_collection if f.file_type.is_image_type)
@@ -99,8 +99,8 @@ class DwmHandler(MooringsHandler):
         is_zip = self.file_type is FileType.ZIP
         have_images = len(images) > 0
         have_netcdfs = len(netcdfs) > 0
-        is_image_zip_pattern = is_zip and DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename)
-        is_calibration_zip_pattern = is_zip and DwmFileClassifier.SOTS_CALIBRATION_ZIP_PATTERN.match(self.file_basename)
+        is_image_zip_pattern = DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename)
+        is_calibration_zip_pattern = DwmFileClassifier.SOTS_CALIBRATION_ZIP_PATTERN.match(self.file_basename)
         zip_file_msg = 'images' if is_image_zip_pattern else 'calibration'
              
         if have_netcdfs:
--- a/aodndata/moorings/handlers.py
+++ b/aodndata/moorings/handlers.py
@@ -99,15 +99,25 @@ class DwmHandler(MooringsHandler):
         is_zip = self.file_type is FileType.ZIP
         have_images = len(images) > 0
         have_netcdfs = len(netcdfs) > 0
-        is_image_zip_pattern = DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename)
-        is_calibration_zip_pattern = DwmFileClassifier.SOTS_CALIBRATION_ZIP_PATTERN.match(self.file_basename)
-        zip_file_msg = 'images' if is_image_zip_pattern else 'calibration'
-             
+        is_image_zip_pattern = have_images and DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename) is not None
+        is_calibration_zip_pattern = DwmFileClassifier.SOTS_CALIBRATION_ZIP_PATTERN.match(self.file_basename) is not None
+        if is_image_zip_pattern:
+            zip_file_msg = 'images'
+        elif is_calibration_zip_pattern:
+            zip_file_msg = 'calibration'
+        else:
+            zip_file_msg = 'no image or calibration file'
+
         if have_netcdfs:
-            raise InvalidFileContentError(
-                "Zip file contains both images or calibration files and netCDFs. Don't know what to do!"
-                " They are handled differently, so please upload only one at a time."
-                )
+            non_netcdfs = self.file_collection.filter_by_attribute_id_not('file_type', FileType.NETCDF)
+            have_only_netcdfs = len(non_netcdfs) == 0
+            if have_only_netcdfs:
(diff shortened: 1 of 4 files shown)


aodndata/moorings/handlers.py
```diff
@@ -93,29 +93,41 @@ def process(self):
...
-                )
-            if not DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename):
-                raise InvalidFileNameError(
-                    "Zip file contains images, but its name does not match pattern for images zip file "
-                    "(regular expression '{p}')".format(p=DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.pattern)
+        is_image_zip_pattern = is_zip and DwmFileClassifier.SOTS_IMAGES_ZIP_PATTERN.match(self.file_basename)
+        is_calibration_zip_pattern = is_zip and DwmFileClassifier.SOTS_CALIBRATION_ZIP_PATTERN.match(self.file_basename)
+        zip_file_msg = 'images' if is_image_zip_pattern else 'calibration'
+             
+        if have_netcdfs:
```
mhidas, line 106: This will raise an error for every valid upload of a zip file containing _only_ netCDFs!

aodndata/moorings/handlers.py
```diff
@@ -93,29 +93,41 @@ def process(self):
...
-                             "Publishing original zip file instead of its contents.")
-
+        elif is_image_zip_pattern or is_calibration_zip_pattern:
+            self.logger.info("Zip file contains {} and no netCDF files. "
+                             "Publishing original zip file instead of its contents.".format(zip_file_msg))
             self.file_collection.set_publish_types(PipelineFilePublishType.NO_ACTION)
             self.input_file_object.publish_type = PipelineFilePublishType.HARVEST_UPLOAD
             self.file_collection.add(self.input_file_object)
+        else:
+            if is_calibration_zip_pattern is None:
```
mhidas, line 118: I think you mean `if not is_calibration_zip_pattern`... but actually this logic is not quite right - it would raise this error for every zip file with images.

test_aodndata/moorings/test_dwmHandler.py
```diff
@@ -13,7 +13,9 @@
 IMAGES_ZIP = os.path.join(TEST_ROOT, IMAGES_ZIP_BASENAME)
 IMAGES_ZIP_CONTENTS = {'saz_2012_47_1000_01.jpg', 'saz_2012_47_1000_02.JPG', 'saz_2012_47_1050_IRS_01.tiff'}
 MIXED_ZIP = os.path.join(TEST_ROOT, 'images_and_nc.zip')
-
+CALIBRATION_ZIP_BASENAME = 'calibration_Pulse-6-2009.zip'
```
mhidas, line 16: We don't really need a 22Mb file for the purpose of this test. It would be nice (non necessary) to make these test files smaller. They can even contain just empty files, since the only thing that matters is their extension.

test_aodndata/moorings/test_dwmHandler.py
```diff
@@ -49,17 +51,42 @@ def test_images_zip_good_names(self):
...
+            handler = self.run_handler_with_exception(InvalidFileNameError, bad_images_zip)
+            #self.assertRegex(handler.error.args[0], r"is an image or calibration zip file but has wrong name")
 
     def test_mixed_zip(self):
         self.run_handler_with_exception(InvalidFileContentError, MIXED_ZIP)
 
+    def test_calib_zip(self):
+        handler = self.run_handler(CALIBRATION_ZIP)
+        assert('calibration' in handler.file_basename)
+        harvested_files = handler.file_collection.filter_by_attribute_id('publish_type', PipelineFilePublishType.HARVEST_UPLOAD)
```
mhidas, line 67: This would be more correct
```suggestion
        harvested_files = handler.file_collection.filter_by_bool_attribute('is_harvested')
```

test_aodndata/moorings/test_dwmHandler.py
```diff
@@ -49,17 +51,42 @@ def test_images_zip_good_names(self):
...
+        assert(len(harvested_files)==1)
+        assert(harvested_files[0].file_type is FileType.ZIP)
```
mhidas, line 69: For consistency with other tests and better integration, please use the `self.assertX' methods instead of just `assert`
```suggestion
        self.assertEqual(len(harvested_files), 1)
        self.assertIs(harvested_files[0].file_type, FileType.ZIP)
```

mhidas (changes requested):
A good start, but we need to discuss the logic.

ana-berger:
Thanks for the review @mhidas. I did some of the changes, and most of them I understood why you suggested it. I might have some questions about a few though. The ones I did not understand I did not change yet.

test_aodndata/moorings/test_dwmHandler.py
```diff
@@ -49,17 +51,42 @@ def test_images_zip_good_names(self):
...
-            self.assertRegex(handler.error.args[0], r"name does not match pattern for images zip file")
+            handler = self.run_handler_with_exception(InvalidFileNameError, bad_images_zip)
+            #self.assertRegex(handler.error.args[0], r"is an image or calibration zip file but has wrong name")
 
     def test_mixed_zip(self):
         self.run_handler_with_exception(InvalidFileContentError, MIXED_ZIP)
 
+    def test_calib_zip(self):
+        handler = self.run_handler(CALIBRATION_ZIP)
+        assert('calibration' in handler.file_basename)
```
ana-berger, line 66: and how do I change this 'assert' here to be consistent. I tried something but didn't do it right

test_aodndata/moorings/test_dwmHandler.py
ana-berger: I was using a 'reduced' version of the file before, but when all the tests were running fine I decided to change to the original file. But I agree, I made the test file smaller now.

ana-berger:
I tried to rearrange the logic, hopefully it is more reasonable now

test_aodndata/moorings/test_dwmHandler.py
```diff
@@ -49,17 +51,42 @@ def test_images_zip_good_names(self):
...
+            self.assertNotEqual(3, len(handler.file_collection))
+            self.assertTrue(handler.input_file_object.is_stored)
+            self.assertTrue(handler.input_file_object.is_harvested)
+
+    def test_calib_zip_bad_names(self):
+        for calib_zip_name in ('NOT_matching_calib_pattern.zip', 'Pulse-6-2009-calib.zip'):
+            bad_calib_zip = os.path.join(self.temp_dir, calib_zip_name)
+            shutil.copy(CALIBRATION_ZIP, bad_calib_zip)
+            handler = self.run_handler_with_exception(InvalidFileNameError, bad_calib_zip)
+            #self.assertRegex(handler.error.args[0], r"is an image or calibration zip file but has wrong name")
```
mhidas, line 85: Again, either update the regex to the expected error message, or delete this line.

mhidas (commented):
Ok, the logic is good now. Just a few more tweaks and some suggestions to simplify the code.

Also, since you added funcionality to `DwmFileClassifier`, ideally you should also test that in test_dwmFileClassifier.py - it will be very similar to the tests for the image zip files (see [here](https://github.com/aodn/python-aodndata/blob/6ea666c8797ef352bcb6e3f7bd29e6d353209e01/test_aodndata/moorings/test_dwmFileClassifier.py#L147))

mhidas (approved):
:+1:

PR #270, opened 2021-06-09
(Fix) add constraint for cftime package for python 3.6

lbesnard:


lbesnard:
@mhidas this fixes the unittests on my end if using a python3.6 env, and hopefully the unittests as well being check by github. I'd like your opinion on this

mhidas (approved):
Ok, that seems to fix the tests.

However, there is actually not much benefit to running these tests in Python 3.6. In prod we're currently on 3.5, and when our various environments are upgraded to Ubuntu 20.04, we'll jump to Python 3.8.

So we need to update our GitHub Actions config to test against 3.5 and 3.8 (as already done in [aodncore](https://github.com/aodn/python-aodncore/blob/b5b2ac978ce410fae4455075cf0b2b7522a9a5e0/.github/workflows/test.yml#L15))

mhidas:
See #271

mhidas:
@lbesnard  I'm happy to also merge this PR if you want to keep using Python 3.6 locally...

constraints.txt
```diff
@@ -1,5 +1,6 @@
 cftime<1.1.1;python_version=='3.5'
+cftime<1.1.1;python_version=='3.6'
 pandas<0.25.0;python_version=='3.5'
 xarray<0.14.0;python_version=='3.5'
 netCDF4<1.5.4;python_version=='3.5'
```
mhidas, line 5: It might be safest to also duplicate this constraint (netCDF4) for python 3.6, as that is the package that will call cftime.

PR #271, opened 2021-06-09
Update GitHub Actions to test against python 3.8

mhidas:
Update the GitHub Actions config to test against the most relevant versions of python, in line with python-aodncore.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,5 +1,5 @@
 cftime<1.1.1;python_version=='3.5'
-cftime<1.5;python_version=='3.8'
+cftime<1.5
 pandas<0.25.0;python_version=='3.5'
 xarray<0.14.0;python_version=='3.5'
 netCDF4<1.5.4;python_version=='3.5'
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,4 +1,5 @@
 cftime<1.1.1;python_version=='3.5'
+cftime<1.5;python_version=='3.8'
 pandas<0.25.0;python_version=='3.5'
 xarray<0.14.0;python_version=='3.5'
 netCDF4<1.5.4;python_version=='3.5'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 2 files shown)


mhidas:
Looks like we have the same problem with `cftime` in 3.8 as in earlier versions (see [error](https://github.com/aodn/python-aodndata/runs/2780455469?check_suite_focus=true#step:6:91))
```
WARNING: The following exceptions occured during the imos:1.3 checker (possibly indicate compliance checker issues):
91
imos:1.3.check_time_coverage: module 'cftime' has no attribute 'utime'
```
None of our code is calling `cftime.utime` directly, so it must be an issue between `netCDF4` and `cftime`. Looks like the `utime` class was removed in cftime 1.5 ([changelog](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/Changelog#L19)).
Let's try pinning it.

constraints.txt
```diff
@@ -1,5 +1,6 @@
 cftime<1.1.1;python_version=='3.5'
+cftime<1.5;python_version=='3.8'
```
lbesnard, line 2: I was actually testing this. probably better to merge the 1st and 2nd line into :
```
cftime<1.5;python_version<='3.8'
```

mhidas: That might break things for 3.5 - I assume there was a reason to pin to `<1.1.1` previously!?

mhidas:
@ndhogan Any thoughts on how to handle issues like this? Do we just keep adding constraints every time something breaks?

constraints.txt
lbesnard: oh my bad, didn't see the diff in version number!

mhidas: But maybe it should be 
```
cftime<1.1.1;python_version=='3.5'
cftime<1.5
```

mhidas:
@lbesnard I reckon we could merge this (instead of your PR)

mhidas:
@digorgonzola FYI - for when you're ready to think about Ubuntu/Python upgrades
(see also https://github.com/aodn/python-aodntools/pull/141 and https://github.com/aodn/cc-plugin-imos/pull/54)

PR #272, opened 2021-06-15
Add conda stage_environmnent.yml file for PO dev

lbesnard:
A conda environment file to simply deploy a working virtual environment. This is based on https://github.com/aodn/python-aodndata/blob/master/scripts/setup_virtualenv.sh

I have had issues with the original script with new version of ubuntu/mint. The old setup_virtualenv.sh script is using the system python to create a virtualenv. On the other end, this conda environment file doesn't rely on this. I would be happy to modify the README of this repo to add an alternative way to PO (like me), who have had issues with the virtualenv script, to setup a working python virtual env.

This doesn't need to be packaged in the wheel, and doesn't affect the repo apart from having a new file. I'm not too fussed if this is not added but I think it would be valuable.

FYI @ana-berger @mhidas @bpasquer @leonardolaiolo

### DIFF
--- a/stage_environment.yml
+++ b/stage_environment.yml
@@ -1,0 +1,28 @@
+# An environment file for Project Officers wanted to use Conda rather than
+# virtualenv
+# This is based on scripts/setup_virtualenv.sh and doesn't interfere with the
+# wheel package
+# STAGE values {rc, build, production}
+# Usage:
+# export STAGE=build && conda env create --file=stage_environment.yml
+# export STAGE=build && conda env update --file=stage_environment.yml
+
+name: python-aodndata-conda
+channels:
+  - conda-forge
+  - defaults
+  - pypi
+dependencies:
+  - cython
+  - cf-units  # force cf-units to be installed from conda-forge. packaged failed otherwise
+  - python=3.8
+  - pip
+  - pip:
+    - numpy<1.19.0
+    - --index-url http://imos-artifacts.s3-website-ap-southeast-2.amazonaws.com/repo/pypi/${STAGE}/
+    - --extra-index-url https://pypi.python.org/simple/
+    - --trusted-host imos-artifacts.s3-website-ap-southeast-2.amazonaws.com
+    - pytest-cov
+    - -c file:constraints.txt
+    - -e .
+    - -e .[testing]


stage_environment.yml
```diff
@@ -0,0 +1,22 @@
...
+# conda env update --file=stage_environment.yml
+name: python-aodndata-conda
+variables:
+  STAGE: production
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - cython
+  - python=3.6
```
mhidas, line 15: Why not 3.8, to match what we'll have in production next?

lbesnard: I was having some compilation issues. I'll try again and put the issues below

lbesnard: @mhidas 
with python 3.8, i get the following error:
```
The conflict is caused by:
    aodntools 1.4.7 depends on pandas<0.25.0 and >=0.24.2
    aodntools 1.4.3 depends on pandas<0.25.0 and >=0.24.2
    aodntools 1.4.1 depends on pandas<0.25.0 and >=0.24.2
    aodntools 1.3.3 depends on pandas<0.25.0 and >=0.24.2
    aodntools 1.3.2 depends on pandas<0.25.0 and >=0.24.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

mhidas: I wonder if this would fix it? https://github.com/aodn/python-aodntools/pull/142

lbesnard:
@mhidas modded the yaml file. not sure if it's the best regarding the environment variable. It's quite a pain to set up. It's easier someone deploys a conda env by running the following
```
export STAGE=build && conda env create --file=stage_environment.yml
```
Alternatively, this could be changed by hard-coding the STAGE value in the yml file directly to only run
```
conda env create --file=stage_environment.yml
```

What to you think is the best, more practical?

lbesnard:
alright, adding cf-units to be installed via conda-forge and not pip fixed it. probably a build issue somewhere.
ready to go!

mhidas (approved):

PR #274, opened 2021-07-19
conditional deployment_id based on frequency

bpasquer:


lbesnard:
would be great to add a unittest for this

bpasquer:
> would be great to add a unittest for this

yes good suggestion , it's done now

PR #276, opened 2021-07-20
Don't pin numpy version in github actions

mhidas:
This should allow unittests to run successfully in both Python 3.5 and 3.8 (see https://github.com/aodn/python-aodncore/pull/235)


mhidas:
@lbesnard @ndhogan as Nate went through the extensive debugging process to figure out how to get these tests to pass for aodncore (https://github.com/aodn/python-aodncore/pull/235), let's apply the same fix here. The tests have been failing on the current master (https://github.com/aodn/python-aodndata/actions/runs/1044336991)

PR #285, opened 2021-10-26
Unittests for ImosBgcDbHandler

mhidas:


mhidas:
@ndhogan I just added a couple more assertions, but I think this is still good to merge.

However, I just realised the GitHub Actions tests have not been triggered for this PR. Any idea why?? (@aodn/devops ??)

mhidas:
> However, I just realised the GitHub Actions tests have not been triggered for this PR. Any idea why?? (@aodn/devops ??)

Same for https://github.com/aodn/python-aodncore/pull/242

ndhogan:
> @ndhogan I just added a couple more assertions, but I think this is still good to merge.
> 
> However, I just realised the GitHub Actions tests have not been triggered for this PR. Any idea why?? (@aodn/devops ??)

@mhidas - GAs are only configured to execute on push / PR into master (see - https://github.com/aodn/python-aodndata/blob/wsd/.github/workflows/test.yml)

mhidas:
> GAs are only configured to execute on push / PR into master

Ah yeah, I missed that point - or rather that we are not yet merging into master here.

mhidas:
@ndhogan When I run this against the updated aodncore (after https://github.com/aodn/python-aodncore/pull/242), my `test_good_csv` now fails because it doesn't have a schema file to validate against.
Any thoughts on the simplest way to patch the config to point to a local schema file?
(Or mock `TableSchemaCheckRunner` to mark the pipelinefile as valid without looking for the schema file?)

mhidas:
This should be good to go now.

PR #287, opened 2021-11-04
Draft PR for Water Sampling DB handler

mhidas:
Tests won't pass until the required [aodncore changes](https://github.com/aodn/python-aodncore/pull/239) are merged.

**TODO**

- [x] Merge & build aodncore
- [x] Update required [aodncore version](https://github.com/aodn/python-aodndata/blob/ca13ca7be41e291bec735adef245e1e605562c35/setup.py#L59)


mhidas:
@ndhogan This should be good to merge now

setup.py
```diff
@@ -11,6 +11,7 @@
         'AuvHandler = aodndata.auv.handler:AuvHandler',
         'ArgoHandler = aodndata.argo.handler:ArgoHandler',
         'AsyncUploadHandler = aodndata.common.asyncupload:AsyncUploadHandler',
+        'ImosBgcDbHandler = aodndata.imos_bgc_db.handler:ImosBgcDbHandler',
```
lbesnard, line 14: is "Imos..." a valid handler name considering that non of the other IMOS handlers have this string in?

mhidas: It's called that because it's a handler for the "IMOS BGC Database" project (aka Octopus, but I resisted the urge to call it that). That's what CSIRO called the project, so I've just stuck with that (the db schema is `imos_bgc_db`)

lbesnard: ok, but then needs to be ordered alphabetically as with other handlers in the list

mhidas: Ah, good point. Done.

ndhogan (approved):

PR #288, opened 2021-11-04
(New) SRS SAR Wind - create new handler - creating plot on the fly

lbesnard:


lbesnard:
@DelphiWard FYI. a new handler which creates a plot on the fly to be attached to the NetCDF.
@leonardolaiolo FYI

No need to merge it yet

lbesnard:
The last two commits remove the dependence to Cartopy and the creation of on-the-fly plots

lbesnard:
https://github.com/aodn/backlog/issues/4197

lbesnard:
The last commits re-revert the changes to re-add cartopy and the creation of plots. it might happen

.github/workflows/test.yml
```diff
@@ -15,22 +15,56 @@ jobs:
...
+        sudo apt-get install -y --no-install-recommends \
+        libproj-dev \
+        libgeos-dev \
+        libffi-dev \
+        cmake \
+        libgtest-dev \
+        build-essential \
+        curl \
+        libcurl4-openssl-dev
+    - name: cartopy - build Gtest needed for proj build
```
digorgonzola, line 39: I think we should use a binary if possible for this so the actions don't take ages

digorgonzola (changes requested):

PR #289, opened 2021-11-11
Srs sar wind stefan

digorgonzola:


=4.9.0
```diff
@@ -0,0 +1 @@
+Looking in indexes: http://imos-artifacts.s3-website-ap-southeast-2.amazonaws.com/repo/pypi/rc/, https://pypi.python.org/simple/
```
lbesnard, line 1: @digorgonzola what is this file? is it supposed to be here?

PR #301, opened 2022-03-11
update required version of testfixtures

evacougnon:
@mhidas this should fix this issue: https://github.com/aodn/content/issues/496
`testfixtures` module is installed in `setup.py` so I didn't update `constraints.txt`

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -4,3 +4,4 @@ pandas<0.25.0;python_version=='3.5'
 xarray<0.14.0;python_version=='3.5'
 netCDF4<1.5.4;python_version=='3.5'
 chardet<4  # temporarily pin to dependency conflict until requests library removes their <4 constraint
+testfixtures<=6.18.3';python_version=='3.5' # >6.18.3 not working under python 3.5.10


mhidas:
Thanks @evacougnon 

> `testfixtures` module is installed in `setup.py` so I didn't update `constraints.txt`

I understand that reasoning, but I don't think `setup.py` is the place to pin that package to a specific version.
The requirements listed in that file are the packages needed for the `aodndata` code to run (sometimes a minimum version is specified if the code relies on functionality introduced in a particular version of a package). 

In this case `testfixtures` is needed to run the unittests, so it's included in the `TESTS_REQUIRE` list. However, the code itself doesn't specifically require version 6.18.3. Any version should do. The reason it's failing is that higher versions of `testfixtures` are incompatible with Python 3.5 . So it's only an issue in the environment we set up to run the code and the tests. The way we've been enforcing such constraints is through the `constraints.txt` file, which is specified in [stage_requirements.txt](https://github.com/aodn/python-aodndata/blob/6c73dc798b4a9e881300ad3d868582e6dddefe64/stage_requirements.txt), and therefore taken into account when the `aodndata` package is installed here:
https://github.com/aodn/python-aodndata/blob/6c73dc798b4a9e881300ad3d868582e6dddefe64/.github/workflows/test.yml#L34

So, long story short, the issue should be fixed by adding a line like this in the constraints.txt file:
```
testfixtures<=6.18.3;python_version=='3.5'
```
(You'll see there are already several similar constraints there, which we'll be able to remove once we've finally got rid of Python 3.5)

evacougnon:
Thanks @mhidas, makes totally sense. I updated the PR, it should be ready once all the tests have passed

mhidas (approved):
:+1: Perfect, thanks!

PR #305, opened 2022-04-12
feat: allow hourly-depth-gridded-product in MooringsFileClassifier

evacougnon:
updated DWM classifier to also work for "hourly-depth-gridded-product": https://github.com/aodn/PO-Backlog/issues/2589


mhidas (approved):
:+1:  Thanks @evacougnon

PR #307, opened 2022-05-10
fix: allow fv03 daily-depth-gridded-product in MooringsFileClassifier

evacougnon:
@mhidas, I'm waiting to have a chat with you before merging. I also need to add a test, so I keep it as a WIP


evacougnon:
closing as it won't be needed anymore

PR #310, opened 2022-05-26
Updating mooring pipeline to ingest NTP profiling mooring files

leonardolaiolo:
@mhidas I believe the check fails are related to the issues that you are already fixing for mooring products? Here is the error I get in my local machine when I run the `test_mooringsProductsHandler.py`:
> FAILED [ 41%]
test_aodndata/moorings/test_mooringsProductsHandler.py:130 (TestMooringsProductsHandler.test_velocity_hourly)
self = <test_aodndata.moorings.test_mooringsProductsHandler.TestMooringsProductsHandler testMethod=test_velocity_hourly>

### DIFF
--- a/aodndata/moorings/classifiers.py
+++ b/aodndata/moorings/classifiers.py
@@ -121,8 +121,10 @@ class MooringsFileClassifier(FileClassifier):
 
         dir_list = [cls.PROJECT]
         dir_list.extend(cls._get_facility(input_file))
-
-        if input_file.endswith('.nc'):
+        if name_fields[1] == 'NTP-PME':
+            dir_list[2] = 'Profiling_Moorings'
+            dir_list.append(cls._get_site_code(input_file))
+        elif input_file.endswith('.nc'):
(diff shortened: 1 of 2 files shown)


mhidas (approved):
:+1: Good work!

mhidas:
> @mhidas I believe the check fails are related to the issues that you are already fixing for mooring products?

Yes, in fact it's a bug I introduced myself :smirk: - it has been fixed in the latest version of aodntools, so all good now!

mhidas:
@leonardolaiolo FYI I was looking at Jenkins and accidentally clicked the `promote_prod` button on aodndata, so this change will be on prod tomorrow. We haven't been able to test it in RC, but I'm pretty confident it won't break anything in prod, so we'll just leave it.

leonardolaiolo:
No worries Marty, thanks for the update. Tomorrow I'll double check and if everything is okay in prod and I can then get in touch with Pete and he can start uploading files.

PR #317, opened 2022-08-08
(Fix) AIMS MMP CTD - change from generic to non generic class to force UPLOAD_ONLY publish type

lbesnard:


lbesnard:
@leo, gotta change the chef-private databag to reflect the new handler name

leonardolaiolo (approved):

PR #327, opened 2022-09-06
4328 setuptools handle runtime version

digorgonzola:


vietnguyengit (dismissed):

aodndata/moorings/burst_average.py
```diff
@@ -25,8 +25,7 @@
 
 from aodntools.ncwriter import ImosTemplate
 from aodndata.moorings.classifiers import MooringsFileClassifier
-from aodndata.version import __version__
-
+from aodndata._version import __version__ as _aodndata_version
```
mhidas, line 28: Would it make sense to hard-code the location of the version string in only one place ([aodncore/__init__.py](https://github.com/aodn/python-aodndata/blob/master/aodndata/__init__.py)), and anywhere else we need it, just import it from the package root?
```suggestion
from aodndata import __version__ as _aodndata_version
```

vietnguyengit (approved):

PR #334, opened 2022-10-05
(Fix) AODN WAVE:

lbesnard:
* aggregation of data could create non strictly monotonic TIME data
   (Fix)
 * implement in handler check on TIME dimension being strictly monotonic
   as the test is not part of the CF 1.7 cc 4.1.1 test suite
 * RT files are archived if they have to be aggregated to an existing
   monthly file. They used to be set to NO_ACTION. Risky if a processing
   needs to re-happen
 * modify unitests to check files against CF 1.7 with a lenient criteria
 * modify unitests to test for aggregation processing in various orders
 * fix some typos


lbesnard:
FYI @mhidas @digorgonzola I had in the last commit to update the standard name vocab in both the github actions and dockerfile, happy to answer some questions. I believe this should be handled differently

lbesnard:
TODO: check dest_path for ARCHIVE data

lbesnard:
> TODO: check dest_path for ARCHIVE data

done

------------

Also fixed some unittests which were running the handler, but testing the dest_path by calling the dest_path static method directly rather than querying the returned value of the dest_path method run within the handler (very different!). Basically, the dest_path values were completely independent from the handler behaviour

ana-berger (approved):

PR #335, opened 2022-12-01
argo: add argo.lock file to INCOMING_DIR during handling process

lbesnard:


lbesnard:
This seems to do the trick, creating a lock file in the incoming dir, and deleting it on postprocessing

mhidas (approved):
:+1: Looks good to me.

The only thing to double-check when you get this on RC is that the handler has the right permissions to _write_ to the incoming dir. 
As an alternative, you could put the lock file in the wip dir where your cron job rsyncs the files to?

mhidas:
Are you ready for me to merge this?

PR #342, opened 2023-01-16
GSLA: update handler with new paths and file names and cleaning code

leonardolaiolo:


lbesnard:
@leonardolaiolo you also need to update (git add, git remove) the test added/removed NetCDF files

leonardolaiolo:
Yes, on it! I guess next time I'll do everything in one go otherwise the tests will keep trying and failing?

leonardolaiolo:
@lbesnard looks like the checks are failing because of the Compliance Checker that does not like that `Conventions global attribute does not contain "CF-1.8"`, do you agree? I think all GSLA NetCDF files have instead `:Conventions = "CF-1.6,IMOS-1.4"`. Thus, do you think this related to the recent Compliance Checker update?

mhidas:
> do you think this related to the recent Compliance Checker update?

Yes. The checker now includes checks against versions 1.6, 1.7 and 1.8 of the CF conventions, and the latest is the default, so you now have to explicitly request `cf:1.6` when you run the checker on any of our files. See https://github.com/aodn/internal-discussions/wiki/IOOS-compliance-checker-issues#cf-17-and-cf-18-checks-now-included

leonardolaiolo:
Thanks @mhidas 
Without updating `cf` to `cf:1.6` I noticed that I had the error raised only here in GitHub and not when I run the unit test in my local.  I was just wondering if this is expected?

mhidas:
> Without updating `cf` to `cf:1.6` I noticed that I had the error raised only here in GitHub and not when I run the unit test in my local. I was just wondering if this is expected?

Probably because you still have the old version of the checker in your local environment.

aodndata/gsla/handler.py
```diff
@@ -13,35 +13,16 @@
 GSLA_REGEX = re.compile(r"""
                        IMOS_OceanCurrent_HV_
                        (?P<nc_time_cov_start>[0-9]{8}T[0-9]{6}Z)_GSLA_FV02_
-                       (?P<product_type>NRT00|DM00|DM01)_C-
-                       (?P<creation_date>[0-9]{8}T[0-9]{6}Z)\.nc(\.gz)?$
+                       (?P<product_type>NRT|DM02)\.nc(\.gz)?$
```
lbesnard, line 16: i would imagine it would be better to have something such as
```python
(?P<product_type>NRT|DM0[1-2])\.nc(\.gz)?$
```
in order to allow multiple versions

aodndata/gsla/handler.py
```diff
@@ -110,27 +89,11 @@ def preprocess(self):
...
 
-        # we don't know the product type (DM00 or DM01) of the file already
-        # on s3 in order to deduce its path. We need to get the product
-        # type from the file in incoming
-        result_previous_version_creation_date = self.get_previous_version_creation_date(netcdf_file.src_path)
-
         """ default values
         by default we push to the storage the file landed in the pipeline (ie *.nc.gz) """
         push_new_file = True
         remove_previous_version = False
```
lbesnard, line 95: might as well remove this since you're not using it anymore

aodndata/gsla/handler.py
```diff
@@ -110,27 +89,11 @@ def preprocess(self):
...
-            new_file_creation_date = get_creation_date(netcdf_file.name)
-            if result_previous_version_creation_date > new_file_creation_date:
-                push_new_file = False
-            elif result_previous_version_creation_date == new_file_creation_date:
-                push_new_file = True
-            else:
-                remove_previous_version = True
-                previous_file_path = self.get_previous_version_object(netcdf_file.src_path)
-
         if push_new_file:
```
lbesnard, line 97: same as above, rewrite this since push_new_file is always true

test_aodndata/gsla/test_gsla.py
```diff
@@ -3,8 +3,9 @@
 import shutil
 import unittest
 from unittest.mock import patch
-from datetime import datetime
+#from datetime import datetime #to use if cannot import TemporaryDirectory
```
lbesnard, line 6: don't commit this comment

test_aodndata/gsla/test_gsla.py
```diff
@@ -43,56 +35,49 @@ def setUp(self):
     def test_push_unzipped_nc(self):
         """ check on pushing *.nc rather than *.nc.gz """
 
-        with TemporaryDirectory() as tmpdir:
-            with gzip.open(GOOD_NC_GZ_DM01, 'rb') as f_in:
-                good_nc_dm01 = os.path.join(tmpdir, os.path.basename(GOOD_NC_GZ_DM01.replace('.gz', '')))
-                with open(good_nc_dm01, 'wb') as f_out:
-                    shutil.copyfileobj(f_in, f_out)
+        #with TemporaryDirectory() as tmpdir:
+        tmpdir =  tempfile.mkdtemp()
```
lbesnard, line 39: don't commit this, use the commented line above and re-indent

test_aodndata/gsla/test_gsla.py
```diff
@@ -184,9 +122,9 @@ def test_setup_upload_location_push_same_file(self):
         # create some PipelineFiles to represent the existing files on 'S3'
```
lbesnard, line 122: this test was there to push the same file again. It kind of mattered because we were checking the creation date in the past. This is not required anymore, so this test should be removed/modified

lbesnard (changes requested):

test_aodndata/gsla/test_gsla.py
```diff
@@ -44,14 +34,14 @@ def test_push_unzipped_nc(self):
         """ check on pushing *.nc rather than *.nc.gz """
 
         with TemporaryDirectory() as tmpdir:
-            with gzip.open(GOOD_NC_GZ_DM01, 'rb') as f_in:
-                good_nc_dm01 = os.path.join(tmpdir, os.path.basename(GOOD_NC_GZ_DM01.replace('.gz', '')))
+            with gzip.open(GOOD_NC_GZ_DM02_nc, 'rb') as f_in:
```
lbesnard, line 37: why not doing this test entirely with the GOOD_NC_GZ_DM02 file instead of GOOD_NC_GZ_DM02_nc. You'll keep one file then. so you can get rid of the variable GOOD_NC_GZ_DM02.nc at the top

leonardolaiolo: Yes, good pick - updated in [commit b90ada8](https://github.com/aodn/python-aodndata/pull/342/commits/b90ada896f5f6a2a5a80fb2ae84209edae66e05f)

test_aodndata/gsla/test_gsla.py
```diff
@@ -84,17 +79,17 @@ def test_bad_file(self):
...
                                         allowed_dest_path_regexes=["IMOS/OceanCurrent/GSLA"])
 
     def test_get_fields(self):
         """ test basic function outputs"""
-        fields = get_pattern_subgroups_from_string(os.path.basename(GOOD_NC_GZ_DM02_nc), GSLA_REGEX)
+        fields = get_pattern_subgroups_from_string(os.path.basename(GOOD_NC_GZ_DM02), GSLA_REGEX)
         self.assertEqual(fields['nc_time_cov_start'], '20000101T000000Z')
         self.assertEqual(fields['product_type'], 'DM02')
 
     def test_good_dm01(self):
```
lbesnard, line 91: rename function name since you're testing dm02

lbesnard (approved):

PR #390, opened 2024-06-12
Publish SOTS gridded multiyear data in new sub-directory

leonardolaiolo:
@mhidas 
I could not use `_get_nc_att` because it was failing some unittest. Indeed, it was raising an `InvalidFileContentError` (you can have a look at my first attempt in commit 435d59c) for the files that do not have 'data_product_type' as attribute. I then avoid using '_get_nc_att' and used instead `_open_nc_file` and `hasattr` to avoid the `InvalidFileContentError` (commit ade1b3f)

### DIFF
--- a/aodndata/moorings/classifiers.py
+++ b/aodndata/moorings/classifiers.py
@@ -326,13 +326,18 @@ class DwmFileClassifier(MooringsFileClassifier):
             dir_list += ['ASFS', 'SOFS', cat, 'Real-time', rt_folder_name]
         elif subfac in ('SOTS', 'ASFS'):
             dir_list.append('SOTS')
-            # Check for data_product_type attribute
-            if cls._get_nc_att(input_file, 'data_product_type'):
-                dir_list.append('derived_products')
-            else:
-                dir_list.append(cls._get_deployment_year(input_file))
-                if cls._is_realtime(input_file):
-                    dir_list.append('real-time')
+            # Open the NetCDF file
+            dataset = cls._open_nc_file(input_file)
+            try:
+                # Check for data_product_type attribute
+                if hasattr(dataset, 'data_product_type'):
+                    dir_list.append('derived_products')
+                else:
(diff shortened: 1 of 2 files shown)


aodndata/moorings/classifiers.py
```diff
@@ -324,9 +326,13 @@ def dest_path(cls, input_file):
             dir_list += ['ASFS', 'SOFS', cat, 'Real-time', rt_folder_name]
         elif subfac in ('SOTS', 'ASFS'):
             dir_list.append('SOTS')
-            dir_list.append(cls._get_deployment_year(input_file))
-            if cls._is_realtime(input_file):
-                dir_list.append('real-time')
+            # Check for data_product_type attribute
+            if cls._get_nc_att(input_file, 'data_product_type'):
```
mhidas: As it is, `cls._get_nc_att` will raise an error if the attribute is missing. What you want here is for it to return `False` in that case, which you can achieve like this:
```python
         if cls._get_nc_att(input_file, 'data_product_type', default=False):
```
See the docstring for `_get_nc_att`: https://github.com/aodn/python-aodncore/blob/d78192d877e3a07d7a8df178cc45ed502a22bb8f/aodncore/pipeline/fileclassifier.py#L75

mhidas: You will also need to save and check the value of the attribute -- see https://github.com/aodn/PO-Backlog/issues/3620#issuecomment-2164241924

mhidas (changes requested):
Thanks @leonardolaiolo . Getting there, but as you will see, we had the requirements slightly wrong (sorry...)

mhidas (approved):

mhidas:
Remember in order to test this in RC you'll need to promote the newly built version of the aodndata pacgate at https://build.aodn.org.au/job/python-aodndata_artifact_promote/

leonardolaiolo:
Good point!

PR #392, opened 2024-07-04
Changed NSW-DPE to NSW in the regex reading file names

Yannick-Rousseau:


bpasquer (changes requested):
The "INSTITUTION_PATHNAME" also needs to be updated; otherwise, the path cannot be set.

lbesnard:
so we agree that the code on data-services reflect his change?

I don't really want to merge this PR right now. 
what is the plan of action for existing files under https://thredds.aodn.org.au/thredds/catalog/Department_of_Planning_and_Environment-New_South_Wales/WAVE-BUOYS/REALTIME/WAVE-PARAMETERS/BENGELLO/2022/catalog.html

If we merge this PR, it means that we can't reprocess existing files if we're not planning on modifying/renaming them. Not ideas

Yannick-Rousseau:
To check with Bene when she is back. 
My understanding is that the department fully wanted to change paths ('directories' names) as well, but this was the compromise.

xhoenner:
Should be good to go now @bpasquer

bpasquer (approved):

PR #403, opened 2025-04-14
Fix-unittests

mphemming:
We have had issues with the unit tests failing for the following PRs:
- https://github.com/aodn/PO-Backlog/issues/3980
- https://github.com/aodn/PO-Backlog/issues/3935
- https://github.com/aodn/backlog/issues/6473

**Error**

The error ([see issue here](https://github.com/aodn/issues/issues/1544)) was due to the `pytz>dev` requirement that `celery==5.1.2` had, which is a [known issue](https://github.com/celery/celery/discussions/9112).

**Cause**
It is a known bug in `celery==5.1.2` that it incorrectly specifies the invalid requirement `pytz>dev`. This is allowed when using `pip==24.0` (as in our case), but if we upgraded pip it would fail.

We can't upgrade celery, because we require `fiona==1.8.18` (as per this requirement `fiona>=1.8.8,<1.8.19` in aodncore) that requires `click<=7.1.2`. Upgrading celery to include the fix requires `click>=8.1.8`...

ChatGPT said it best: "You’re in a tough corner of dependency hell"

**Solution**

During the Github Action 'Install dependencies' step, the celery package is downloaded, the requirements files are edited to use `pytz>=2020.1`, and then the package is installed. This bit removes the `pytz>dev` dependency that was causing trouble.  This seemed to be the only option as upgrading celery would cause conflicts with the other packages.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,9 @@ jobs:
         pip install scipy
         pip install -r stage_requirements.txt
         pip install pytest-cov
+        pip install click==7.1.2
+        pip install celery==5.1.2
+        pip install fiona==1.8.18
       env:
         STAGE: ${{ matrix.stage }}
     - name: Test with pytest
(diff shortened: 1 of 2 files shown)


leonardolaiolo (approved):
It looks like the test are running succesfully

PR #405, opened 2025-06-16
Fix dependencies issue

mphemming:
This was a bit of a mess to clean up but it is now working. It is only a band-aid solution for GH Actions. I don't have much knowledge about how `python-adondata` works to get stuck into exploring and testing the repo. And I haven't touched the requirements for the repository, hence there is now quite a difference between the packages inside the container used for the GH Actions and the packages used for the repository (see below). 

### Initial error

`× Getting requirements to build wheel did not run successfully.` (see full error [here](https://github.com/aodn/issues/issues/1560#issue-3148168777))

This error occured when trying to install `shapely==2.0.7` by default, and not finding `np.intp_t` in the version of Numpy that was available, hence the build crashes.  

### Solution

- pin safe package versions: `numpy<1.24.0` -> contains `np.intp_t`, `pandas<2.0` to avoid any breaking changes, and `matplotlib>=3.1,<3.6` that works well with `Cartopy 0.20.3` (another requirement). 
- ensure `cartopy` and `shapely` have access to their C libraries (GEOS, PROJ, GDAL/OGR, UDUNITS2). These libraries are installed and available in the container.
- pin the last safe version of `shapely` (`1.8.5.post1`) that will work with `numpy<1.24.0` and the C libraries.
- rebuild `cartopy=0.20.31` from source (as not available as a wheel), linking to C libraries.

> Note: I am not using stage_requirements.txt at all now in the GH Actions

### Questions

- As we are shifting to the new pipeline, is it sufficient to fix the GH Actions only? or should we test and update the requirements in the repository as well?
- Should we still be using stage_requirements.txt?

### Different environments

| Package                | GH Actions Version | Local Version     |
|:-----------------------|:-------------------|:------------------|
| bitarray               | 3.4.3              | 3.0.0             |
| bitstring              | 4.3.1              | 4.3.0             |
| boto3                  | 1.37.38            | 1.37.17           |
| botocore               | 1.37.38            | 1.37.17           |
| Cartopy                | 0.20.3             | 0.21.1            |
| certifi                | 2025.6.15          | 2025.1.31         |
| cftime                 | 1.6.4.post1        | 1.4.1             |
| chardet                | 5.2.0              | 3.0.4             |
| charset-normalizer     | 3.4.2              | 3.4.1             |
| click-plugins          | 1.1.1.2            | 1.1.1             |
| coverage               | 7.6.1              | (not installed)   |
| cryptography           | 45.0.5             | 44.0.2            |
| exceptiongroup         | 1.3.0              | (not installed)   |
| fonttools              | 4.57.0             | 4.56.0            |
| httpretty              | 1.1.4              | (not installed)   |
| importlib_metadata     | 8.5.0              | (not installed)   |
| iniconfig              | 2.1.0              | (not installed)   |
| kombu                  | 5.5.4              | 5.5.0             |
| lxml                   | 6.0.0              | 5.3.1             |
| matplotlib             | 3.5.3              | 3.7.5             |
| packaging              | 25.0               | 24.2              |
| prompt_toolkit         | 3.0.51             | 3.0.50            |
| pybufrkit              | 0.2.25             | 0.2.24            |
| pytz                   | 2025.2             | 2025.1            |
| requests               | 2.32.4             | 2.32.3            |
| s3transfer             | 0.11.5             | 0.11.4            |
| scipy                  | 1.10.1             | (not installed)   |
| setuptools-scm         | 8.3.1              | (not installed)   |
| Shapely                | 1.8.5.post1        | 2.0.7             |
| SQLAlchemy             | 2.0.41             | 2.0.39            |
| testfixtures           | 8.3.0              | (not installed)   |
| tomli                  | 2.2.1              | (not installed)   |
| transitions            | 0.9.3              | 0.9.2             |
| xlrd                   | 2.0.2              | 2.0.1             |

FYI @bpasquer @vfisaac


.github/workflows/test.yml
```diff
@@ -28,46 +28,62 @@ jobs:
...
+    - name: Install dependencies (inline, no external reqs)
       run: |
-        python -m pip install --upgrade pip==24.0
+        # 1. Pin pip & setuptools so distutils & flags work
+        python -m pip install --upgrade pip==24.0 setuptools==65.5.1 wheel setuptools_scm
 
-        # patch to solve dependency issues with celery / fiona / click / pytz
-        # download celery package 
+        # 2. Point pip at your IMOS PyPI repo (so aodntools>=1.3.1 is found)
+        export PIP_INDEX_URL="http://imos-artifacts.s3-website-ap-southeast-2.amazonaws.com/repo/pypi/${STAGE}/"
```
lbesnard, line 54: is this wookring? STAGE is an VENV, but then "stage" is defined above. not the same case?

lbesnard (approved):

akashisama:
Since this only touches the github action code it won't break legacy pipelines. Any changes to fix dependencies outside of github actions (eg stage-requirements.txt) though could break legacy, so I think the best policy is to not touch the stuff outside actions until we're either forced to do so by something breaking or we've decommissioned the legacy pipelines

akashisama (approved):

mphemming:
@bpasquer @vfisaac this PR has now been merged

bpasquer:
[celebrate]     Benedicte Pasquer reacted to your message:
________________________________

PR #406, opened 2025-08-05
Feat: surface waves altimetry - add SWOT

lbesnard:


lbesnard:
@leonardolaiolo , sorry , there is only one meaningful change with was to add the SWOT sat. the rest was some auto indent fix

leonardolaiolo:
@lbesnard just double checking here, it looks like there are some failings due to dependencies - this is not related to the changes in  `srs-surface-waves`, do you wna to go ahead anyway?

lbesnard:
@leonardolaiolo this is now rebased, so good to go

leonardolaiolo (approved):

